PR #133, opened 2015-04-01
Vtk6 support

jcfr:
@lantiga This allows the `SlicerExtension-VMTK` to compile. More work is needed to update the python scripts of both the `VTMK` and `SlicerExtension-VMTK` projects to work with both VTKv5 and VTKv6.

To faciliate the work, publishing your test suite and data would be great. Doing so would ensure there are no regressions and will increase the confidence to accept contributions.

Note(s):
* The code is expected to compile against VTK 5.10 but I didn't try it. I can only confirm it compiles against VTK 6.2

Ping @pieper  @haehn


lantiga:
Wow, amazing JC!

I need to see if it works against VTK 5.10 (our fork) before merging to master. In any case the intention is to transition to VTK 6 as soon as we complete the port of the python components.
Let me know if you want to retarget the pull request on a branch other than master (vtk6-support is fine), so I can merge it right away. Alternatively I'll keep the pull request open until I check that 5.10 still works.

As for making tests public I have two issues right now. One is data, which come from previous projects and it is not mine to distribute. I need to settle on distributable data and convert tests scripts to use it. The other is that I do regression testing from vmtkScripts, while I'd like to test individual classes from ctest. But I agree that one way or another we need to resolve this ASAP.

Thank you again for the great contribution.

lantiga:
Merged. With an extra commit things will work on VTK5.10, including the Python part. Thanks a lot again.

jcfr:
:+1:

PR #163, opened 2016-03-10
Make homebrew file work for non-/usr/local prefix

rupertnash:
The VMTK homebrew formula only works if brew is installed in /usr/local.

This patch makes compatible with non-standard locations.


lantiga:
Excellent, thanks

PR #164, opened 2016-03-30
Homebrew dependency tweaks

rupertnash:
In the course of getting VMTK built with homebrew, I changed a few things that you may wish to merge into master

1. VMTK doesn't require that you use the homebrew python (I have successfully built it using anaconda with the amended formula). To this end, I have set PYTHON_EXECUTABLE using `which python`.

2. Python is a run time requirement as well as build time (unless you turn off python wrapping and the Pypes stuff, I guess - but this isn't implemented in the Formula)

3. ITK is a runtime dependancy as well as build time

4. VTK automatically includes python. 

Cheers,
Rupert

### DIFF
--- a/distribution/homebrew/vmtk.rb
+++ b/distribution/homebrew/vmtk.rb
@@ -14,6 +14,7 @@ class Vmtk < Formula
   depends_on "cmake" => :build
   depends_on "vtk"
   depends_on "insightToolkit"
+  depends_on :python
   
   def install
     args = std_cmake_args + %W[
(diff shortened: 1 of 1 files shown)


lantiga:
Thank you Rupert, sorry for the late reaction. This is great, installation is a breeze now.

Eventually this vtk issue (https://github.com/Homebrew/homebrew-science/issues/3401) will get fixed, so things will work with brewed python as well.
Tomorrow we'll change the instructions on the website, thanks again.

gnamor:
Dear Luca, 
I'm looking for a solution to the vtk issue (Homebrew/homebrew-science#3401) you mentioned.
In your reply above it sounds like the issues is going to be solved, if not already solved.
Is it?
I just installed the latest VTK with brew and the problem is still there.
Thank you,
Roman

PR #166, opened 2016-05-10
Added the missing 'vmtkdistancetospheres' in vmtkscripts.py

ChrisChnafa:


lantiga:
`vmtkdistancetospheres` is already listed in `vmtk/vmtkScripts/contrib/vmtkcontribscripts.py`
You can include the script with
`from vmtkcontribscripts import vmtkdistancetospheres`

PR #168, opened 2016-09-22
Added functionality to change the position of a window

wouterpotters:
Added functionality to change the position of a window (useful when comparing several pipelines)

Previously this worked:
`self.vmtkRenderer.WindowSize = [sizex, sizey]`
`-position sizex sizey`

Now this works as well:
`self.vmtkRenderer.WindowPosition =  [posx, posy]`
`-position posx posy`


lantiga:
Thank you!

PR #171, opened 2016-10-07
Update VMTK to be compatible with VTK7

lassoan:
Slicer4 uses VTK7 and therefore VMTK had to be updated to support VTK7, too.

All the changes should be backward-compatible, so VMTK works the same way for older VTK versions.

Python scripts have not been updated for VTK7. There should have been a lot of modifications because of SetInput->SetInputData changes. It is doable, see example below, I just did not need it and if backward compatibility is not needed then the changes would be much simpler.

    if vtk.VTK_MAJOR_VERSION <= 5:
      somefilter.SetInput(self.brush)
    else:
      somefilter.SetInputData(self.brush)

### DIFF
--- a/vtkVmtk/ComputationalGeometry/vtkvmtkMergeCenterlines.cxx
+++ b/vtkVmtk/ComputationalGeometry/vtkvmtkMergeCenterlines.cxx
@@ -315,7 +315,11 @@ int vtkvmtkMergeCenterlines::RequestData(vtkInformation *vtkNotUsed(request), vt
   tupleValue[0] = tupleValue[1] = -1;
   for (i=0; i<numberOfMergedCells; i++)
     {
+#if VTK_MAJOR_VERSION < 7
     cellAdditionalEndPointIds->SetTupleValue(i,tupleValue);
+#else
+    cellAdditionalEndPointIds->SetTypedTuple(i, tupleValue);
+#endif
     }
 
   for (i=0; i<blankedGroupIds->GetNumberOfIds(); i++)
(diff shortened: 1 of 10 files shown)


lantiga:
Hi Andras, thank you! We're actually compiling vmtk against VTK 7.0.0 (see SuperBuild.cmake), maybe we were relying on older/deprecated signatures? I'll gladly merge the new ones.
Python scripts have already been ported to VTK > 5 (we ditched backward compatibility a while ago for those).
Thanks again!

lassoan:
Maybe VTK_LEGACY_REMOVE flag is defined in Slicer's VTK build but not in VMTK's.

lassoan:
It would be nice if you could test SetUpdateExtent changes, as I just copied the changes that were made in similar VTK classes in Slicer.

lantiga:
I see, that's it: we don't define VTK_LEGACY_REMOVE.
Thanks for the heads up.

mobernabeu:
Following 36ee6b4, I'm getting the following missing header error when compiling VMTK from source (I let VMTK install its own VTK/ITK).
```
/tmp/vmtk/vtkVmtk/Utilities/vtkvmtkITK/vtkvmtkITKArchetypeImageSeriesScalarReader.cxx:23:37: fatal error: vtkAOSDataArrayTemplate.h: No such file or directory
```
Is this something you can reproduce on your end?

lassoan:
Thanks for testing. The commit that fixed compatibility with more recent VTK versions had inaccurate version checks for generic array related changes (generic data arrays were introduced in VTK 7.1 and the checks were for VTK 7 only, without checking the minor version).
A pull request has been submitted with the fix (https://github.com/vmtk/vmtk/pull/172).

mobernabeu:
Thanks @lassoan! #172 got the code compiling again on my end.

PR #172, opened 2016-10-11
COMP: Made VMTK compatible with both VTK 7.0 and 7.1 (and later)

lassoan:
Generic data array was only introduced in VTK7.1 and not in VTK7. Adjusted #ifdefs to build correctly with both VTK 7.1 and VTK7.1 (and later).


lantiga:
Thank you!

PR #174, opened 2017-02-26
Python 3 Support

rlizzo:
Pull request to discuss changes required for vmtk to support python 3. Source is currently building against a clean python 3.5 install (via anaconda virtual environment). Initial version of pull request will only build against python 3.5 (as install directory is currently hardcoded against it). 

High level topics of discussion include:

- Python 2 - Python 3 backwards compatibility.
- Distribution and packaging while ensuring that relative imports remain in place.


PypeS/pype.py
```diff
@@ -236,20 +238,18 @@ def Execute(self):
...
+                self.PrintError(str(e))
                 break
-            scriptObjectClassName = ''
-            exec ('scriptObjectClassName =  '+moduleName+'.'+moduleName)
-            moduleScriptObjectClassName = moduleName+'.'+scriptObjectClassName
-            self.PrintLog('Creating ' + scriptObjectClassName + ' instance.')
-            scriptObject = 0
-            exec ('scriptObject = '+moduleScriptObjectClassName+'()')
+            scriptObject = getattr(module, scriptObjectClassName)
+            scriptObject = scriptObject()
```
rizzol: I wanted to highligh this section (and the import section directly above) since this is a major change to the stucture of the code. 

The combination of lines 244-249 and 251-252 decouples the explicit definition in the first line of a vmtkScript between the files name and the name of the class inside. Instead, the scriptObjectClassName is identified by searching attributes with a file and selecting the first class that it comes across, dynamically creating the scriptObject from a call to that scriptObjectClassName

lantiga: I'm all for this.

The only potential issue is getting the first class that comes across. There are script modules that contain multiple classes (e.g. vmtkcenterlines.py) that would violate the assumption.

Instead, we should get the first class that is a descendant of `pypes.pypeScript`, making the reasonable assumption that there would be just one per module.

The would remove the need for having a line like
`vmtkcenterlines = 'vmtkCenterlines'`
in every script, which was a total hack. The code assumes that there is one variable in the module named as the module (i.e. `vmtkcenterlines`), whose value is the name of the class. The `exec ('scriptObjectClassName =  '+moduleName+'.'+moduleName)` therefore assigns the name of the class to the `scriptObjectClassName`.
With your change, we should also remove all such variables from scripts.

PypeS/pypepad.py
```diff
@@ -29,7 +29,7 @@ def __init__(self,tk_text_widget):
         self.output_file = None
   
     def write(self,text):
-        from Tkinter import NORMAL, END, DISABLED
+        from tkinter import NORMAL, END, DISABLED
```
lantiga: Thinking about the `Tkinter` -> `tkinter` change, we could simply

PypeS/pypeserver.py
```diff
@@ -36,15 +37,15 @@ def RunPypeProcess(arguments, inputStream=None, outputStream=None, logOn=True):
         pipe.OutputStream = outputStream
     pipe.LogOn = logOn
     pipe.LogOn = True
-    if type(arguments) in [str,unicode]:
+    if type(arguments) in [str, str]:
```
lantiga: `if type(arguments) == str:`

vmtkApps/CerebralAneurysms/ParentVesselReconstruction/clipvoronoidiagram.py
```diff
@@ -270,13 +272,13 @@ def SmoothClippedVoronoiDiagram(voronoi,centerlines):
...
 
-print 'Clipping Voronoi Diagram'
+print('Clipping Voronoi Diagram')
 maskedVoronoiArray    = MaskVoronoiDiagram(originalVoronoiDiagram,patchCenterlines)
 clippedVoronoiDiagram = ExtractMaskedVoronoiPoints(originalVoronoiDiagram,maskedVoronoiArray)
 WritePolyData(clippedVoronoiDiagram, clippedVoronoiDiagramFilename)
 
 if (smoothVoronoiDiagram==1):
-   print 'Smoothing Voronoi Diagram',
+   print('Smoothing Voronoi Diagram', end=' ') # end seperator new in py3, will throw error in py2 environments
```
lantiga: We don't need `end=' '` here. Line 279 in the original code just had a training comma.

lantiga:
I inspected the code, everything looks good to me. 
Apart from a couple of minor comments, I added a comment here https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-0e61c56d3c5319df5327ad0822c38c30R252 that should lead to further simplification.

Just to keep track, here's what we should do to make things work in python2.x as well as 3.x:
- [ ] make all occurrences of `python3.5` pick the right version from a CMake variable in CMakeLists.txt (both in vmtk base and in vtkVmtk), as well as `vmtk_env.sh.in`
- [ ] import Tkinter on python2.7, something like the following should suffice
   ```
   try:
       import tkinter
       import tkinter.filedialog as filedialog
       from tkinter import Tk
   except:
       import Tkinter as tkinter
       import tkFileDialog as filedialog
       from Tkinter import Tk
   ```
   See also all uses of `tkFileDialog` to handle the `BROWSER` argument to `-ifile`, `-ofile` etc. 
- [ ] revamp [vmtk.rb](https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-37ed75bfd1cb7bceab13fbb4098c6048). The way it is it searches for python using `which python`, which will typically pick up python2.x. Let's use the `vtk.rb` recipe here https://github.com/Homebrew/homebrew-science/blob/master/vtk.rb#L64 as reference.
- [ ] make definition of `PYTHONPATH` in vmtk.py conditional wrt the version
- [ ] check that `input` works instead of `raw_input` in python2.x as well, see here https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-42c6316d7b89b68b4dfdaf08abcb0735R83
- [ ] tests!

@rlizzo, I created a `python3` branch. Could you please change your target to that branch, so that we can merge in the near future, and then work at the 2+3 fixes separately? 

Thanks! 👍🏻

lantiga:
Other point to fix:
- [ ] generate correct shebang (e.g. `/usr/local/bin/python3` vs `/usr/local/bin/python`) according to the interpreter we build against

lantiga:
I have an issue with local imports, e.g. `from . import pypes`.
```
Unexpected error: <class 'ImportError'>
Traceback (most recent call last):
  File "/Users/lantiga/Desktop/vmtk-python3/vmtk-build/Install/bin/vmtkimageviewer", line 20, in <module>
    import pypeserver
ImportError: No module named 'pypeserver'
```
I'm calling `vmtkimageviewer --help` from outside the build directory, after I properly configured environment variables
```
VMTKHOME=/Users/lantiga/Desktop/vmtk-python3/vmtk-build/Install
export PATH=$VMTKHOME/bin:$PATH
export DYLD_LIBRARY_PATH=$VMTKHOME/lib:$DYLD_LIBRARY_PATH
export PYTHONPATH=$VMTKHOME/lib/python3.5/site-packages:$PYTHONPATH
```
and manually changed the shebang.

However, changing the above to `from vmtk import pypes` works for me in python3 as it did in python2.

lantiga:
(Sorry for the multiple comments, I'm writing as I go)
I'm trying to execute `vmtkimageviewer -ifile foo.vti`, I get an error related to 
`exec('filename = self.' + self.GetIOInputFileNameMember(member.MemberName))`, since `exec` is now a function and the local scope will essentially not see `filename` anymore.
Virtually every such situation can be turned into 
`filename = eval('self.' + self.GetIOInputFileNameMember(member.MemberName))`
which works and is arguably cleaner.

Probably there are cleaner ways, but this is a straightforward change and it works in 2+3.

BTW, we can merge into the `python3` branch and then fix these issues collaboratively.

rlizzo:
Thanks luca, will respond to your comments in detail, but just target branch to vmtk/python3 and am about to merge now.

PR #175, opened 2017-03-03
Python3

rlizzo:
I'm creating a second pull request from vmtk/python3 to vmtk/master so to keep track of changes as they are made and keep the comments in a recordable form. I'm copying your feedback from the initial pull request to this one.


PypeS/pypepad.py
```diff
@@ -276,7 +276,7 @@ def GetSuggestionsList(self,word):
             scriptindex = self.text_input.search('vmtk',self.wordIndex[0],backwards=1)
             moduleName  = self.text_input.get( scriptindex,scriptindex+' wordend' )
             try:
-                exec('import '+moduleName)
+                exec('from . import '+moduleName)
                 exec('scriptObjectClassName =  '+moduleName+'.'+moduleName)
```
rlizzo: I wanted to highlight this section (and the import section directly above) since **this is a major change to the structure of the code.** 

The combination of lines 244-249 and 251-252 decouples the explicit definition in the first line of a vmtkScript between the files name and the name of the class inside. Instead, the scriptObjectClassName is identified by searching attributes within a file and selecting the first class that it comes across, dynamically creating the scriptObject from a call to `getattr(module, scriptObjectClassName)()
`
The benefits of this change are:

1. This is a much **more "pythonic"** way of handling dynamic object imports. It takes advantage of the core language features (everything is a first-class object, and self introspection) which make python so powerful for the wide variety of use cases. 
2. It is much **more transparent and readable** - this codeblock is a core method of the library, and the use of exec() obfuscates the meaning and importance of the section. 
3. We should see **significant performance improvements** over the analogous `exec()` call. The reasons are two-fold:
    1. In the modified form, the interpreter has all the required variables available in it's local scope, and is able to optimize execution of the module import and class instantiation . This is in direct contrast to the full-halt enabled whenever an `exec() `call is made, a process where the interpreter has to compile the import string and any associated object to bytecode (a time consuming process) before actually doing any of the loading and class instantiation in an isolated scope whose objects must be backported into the calling functions locals() / globals() namespace. 
    2. When executing a function within the scope of an` exec() `call, the (cyclicly executing) python garbage collector cannot identify the scope which any objects created within the  `exec() `namespace were created in. Effectively, this means that the garbage collector does not run on these objects! Objects created within an exec() call have to explicitly call the` __del__ `deconstructor, or the garbage collector will not remove the objects from memory. 

_Preferably, we would remove all uses of the exec() function from the library._ There are very few use-cases of exec which cannot be more optimally and more transparently executed with python standard library functions. 

More info at the following links:
- http://lucumr.pocoo.org/2011/2/1/exec-in-python/
- http://stackoverflow.com/a/29456463
- https://late.am/post/2012/04/30/the-exec-statement-and-a-python-mystery.html

rlizzo: ------
@lantiga  Reply
-----

I'm all for this.

The only potential issue is getting the first class that comes across. There are script modules that contain multiple classes (e.g. vmtkcenterlines.py) that would violate the assumption.

Instead, we should get the first class that is a descendant of `pypes.pypeScript`, making the reasonable assumption that there would be just one per module.

The would remove the need for having a line like
`vmtkcenterlines = 'vmtkCenterlines'`
in every script, which was a total hack. The code assumes that there is one variable in the module named as the module (i.e. `vmtkcenterlines`), whose value is the name of the class. The `exec ('scriptObjectClassName =  '+moduleName+'.'+moduleName)` therefore assigns the name of the class to the `scriptObjectClassName`.
With your change, we should also remove all such variables from scripts.

PypeS/pypeserver.py
```diff
@@ -36,15 +37,15 @@ def RunPypeProcess(arguments, inputStream=None, outputStream=None, logOn=True):
         pipe.OutputStream = outputStream
     pipe.LogOn = logOn
     pipe.LogOn = True
-    if type(arguments) in [str,unicode]:
+    if type(arguments) in [str, str]:
```
rlizzo: ---
@lantiga reply
---

`if type(arguments) == str:`

vmtkApps/CerebralAneurysms/ParentVesselReconstruction/clipvoronoidiagram.py
```diff
@@ -270,13 +272,13 @@ def SmoothClippedVoronoiDiagram(voronoi,centerlines):
...
 
-print 'Clipping Voronoi Diagram'
+print('Clipping Voronoi Diagram')
 maskedVoronoiArray    = MaskVoronoiDiagram(originalVoronoiDiagram,patchCenterlines)
 clippedVoronoiDiagram = ExtractMaskedVoronoiPoints(originalVoronoiDiagram,maskedVoronoiArray)
 WritePolyData(clippedVoronoiDiagram, clippedVoronoiDiagramFilename)
 
 if (smoothVoronoiDiagram==1):
-   print 'Smoothing Voronoi Diagram',
+   print('Smoothing Voronoi Diagram', end=' ') # end seperator new in py3, will throw error in py2 environments
```
rlizzo: ---
@lantiga reply
---

  Preview

We don't need `end=' '` here. Line 279 in the original code just had a training comma.

rlizzo:
---
@lantiga reply
---

I inspected the code, everything looks good to me. 
Apart from a couple of minor comments, I added a comment here https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-0e61c56d3c5319df5327ad0822c38c30R252 that should lead to further simplification.

Just to keep track, here's what we should do to make things work in python2.x as well as 3.x:
- [ ] make all occurrences of `python3.5` pick the right version from a CMake variable in CMakeLists.txt (both in vmtk base and in vtkVmtk), as well as `vmtk_env.sh.in`
- [ ] import Tkinter on python2.7, something like the following should suffice
   ```
   try:
       import tkinter
       import tkinter.filedialog as filedialog
       from tkinter import Tk
   except:
       import Tkinter as tkinter
       import tkFileDialog as filedialog
       from Tkinter import Tk
   ```
   See also all uses of `tkFileDialog` to handle the `BROWSER` argument to `-ifile`, `-ofile` etc. 
- [ ] revamp [vmtk.rb](https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-37ed75bfd1cb7bceab13fbb4098c6048). The way it is it searches for python using `which python`, which will typically pick up python2.x. Let's use the `vtk.rb` recipe here https://github.com/Homebrew/homebrew-science/blob/master/vtk.rb#L64 as reference.
- [ ] make definition of `PYTHONPATH` in vmtk.py conditional wrt the version
- [ ] check that `input` works instead of `raw_input` in python2.x as well, see here https://github.com/vmtk/vmtk/pull/174/files/452d61c434b315d194a5e186a7816bea7d9d1c0c#diff-42c6316d7b89b68b4dfdaf08abcb0735R83
- [ ] tests!

@rlizzo, I created a `python3` branch. Could you please change your target to that branch, so that we can merge in the near future, and then work at the 2+3 fixes separately? 

Thanks! 👍🏻

rlizzo:
---
@lantiga reply
---

Other point to fix:
- [ ] generate correct shebang (e.g. `/usr/local/bin/python3` vs `/usr/local/bin/python`) according to the interpreter we build against

rlizzo:
---
@lantiga reply
---

I have an issue with local imports, e.g. `from . import pypes`.
```
Unexpected error: <class 'ImportError'>
Traceback (most recent call last):
  File "/Users/lantiga/Desktop/vmtk-python3/vmtk-build/Install/bin/vmtkimageviewer", line 20, in <module>
    import pypeserver
ImportError: No module named 'pypeserver'
```
I'm calling `vmtkimageviewer --help` from outside the build directory, after I properly configured environment variables
```
VMTKHOME=/Users/lantiga/Desktop/vmtk-python3/vmtk-build/Install
export PATH=$VMTKHOME/bin:$PATH
export DYLD_LIBRARY_PATH=$VMTKHOME/lib:$DYLD_LIBRARY_PATH
export PYTHONPATH=$VMTKHOME/lib/python3.5/site-packages:$PYTHONPATH
```
and manually changed the shebang.

However, changing the above to `from vmtk import pypes` works for me in python3 as it did in python2.

rlizzo:
---
@lantiga reply
---

(Sorry for the multiple comments, I'm writing as I go)
I'm trying to execute `vmtkimageviewer -ifile foo.vti`, I get an error related to 
`exec('filename = self.' + self.GetIOInputFileNameMember(member.MemberName))`, since `exec` is now a function and the local scope will essentially not see `filename` anymore.
Virtually every such situation can be turned into 
`filename = eval('self.' + self.GetIOInputFileNameMember(member.MemberName))`
which works and is arguably cleaner. Probably there are cleaner ways, but this is a straightforward change and it works in 2+3.

After this and other 2 analogous changes in `pypescript.py`, the viewer works correctly 🎉.

BTW, we can merge into the `python3` branch and then fix these issues collaboratively.

rlizzo:
Thanks luca, will respond to your comments in detail, but just changed the target branch to vmtk/python3 and am about to merge now. 

---

note: am only including this for the sake of completeness. This pull request is in the vmtk/python3 branch

vmtkApps/CerebralAneurysms/ParentVesselReconstruction/clipvoronoidiagram.py
rlizzo: fixed this in latest push to the branch

PypeS/pypeserver.py
rlizzo: fixed this in latest push to this branch.

PypeS/pypepad.py
rlizzo: OK. I'm going to try something like: `assert( issubclass(scriptObjectClassName, pypes.pypeScript)) ` and see how that works - it should solve the problem, but i can't help but feel like this is a wierd workaround. 

Out of curiosity, are you primarily a Java programmer? This library has some distinctive java-esque influence on it. It's not a bad thing, but it's been a bit of a challenge for me to adapt my coding style to understand it :+1:

rlizzo:
In response to your comment 

>  Other point to fix:
> - [ ] generate correct shebang (e.g. `/usr/local/bin/python3` vs `/usr/local/bin/python`) according to the interpreter we build against

I modified the scripts in the PypesS module to read `#!/usr/bin/env python` to take care of this. This method allows the OS to choose the current path of the python install to use, which make it very portable (especially when building against virtual environments). Shouldn't be an issue if we change everything to the new form.

rlizzo:
I fixed the relative imports problem by renaming all `from . import` to `from vmtk import`. This should fix some of the import issues.

lantiga:
Hi Rick, I had a hell of a week. Sincere apologies. I'm back, I'll resume my comments now.

lantiga:
Regarding 
```
#!/usr/bin/env python
```
it's a long story. It was that way once, we changed it to `#!${PYTHON_SHEBANG}`, i.e. we have CMake generate it. If you look in `CMakeLists.txt` in the base directory, you'll see that `PYTHON_SHEBANG` is 
```
SET(PYTHON_SHEBANG ${PYTHON_EXECUTABLE})
```
on the Mac, while `/usr/bin/env python` on Linux.

Here's the catch: for vmtk to be imported, we need to modify `DYLD_LIBRARY_PATH`. Such env variable is typically inherited by the process you invoke. However, since macOS Sierra, `/usr/bin/env` wipes out that env var for security reasons before calling the argument (`python` in this case).

The only hacky workaround we could find was to have CMake generate the shebang based on the value of the `PYTHON_EXECUTABLE` CMake variable (on macOS only). If that is python 3, then python 3 is used.

PypeS/pypepad.py
lantiga: Ahah :-) This library was one of the first things I wrote in Python a loong time ago. At the time I had just learned object oriented programming from a C++ book, maybe that's what you're picking up. Or maybe it's just weird in its own self 😄 

My suggestion here is to change:
```
    scriptObjectClasses = [x for x in dir(module) if isclass(getattr(module, x))]
```
to
```
    scriptObjectClasses = [x for x in dir(module) if isclass(getattr(module, x)) and issubclass(getattr(module, x), pypes.pypeScript)]
```
This would guarantee we don't bump into some extra classes defined in the module when we get `scriptObjectClassName = scriptObjectClasses[0]`.

rlizzo:
Thanks for the info Luca! 

I'm not a Mac guy myself; Since you have a long history with this problem (and since it's not a particular problem in Windows or Linux) is it OK if I leave the `shebang` issue to you?

The other option that may be worth considering is to drop support for python 2 all together. Making the full transition to python 3 would be significantly simpler, and allow us to take advantage of new language features and runtime optimisations. Besides, the end of development support for python 2 is [not too far away](https://pythonclock.org/). 

Anyways, since I can't see these changes being merged into `master` without identifying them as a new VMTK version, is there any reason we can't drop support for python 2? 

I know it's a pretty big decision to make, so no rush. We can proceed as if we wanted full backwards compatibility until you let me know otherwise - I'll get behind any decision you make.

rlizzo:
Hey luca, 

I fixed some of the import file errors as per your suggestion to change exec to eval. I also changed python versions from 3.5 to 3.6 since vtk7.1 works with python3.6. 

I'm now able to read in a file, do a level set segmentation, marching cubes, surface viewer, surface writer. It's comming along.

lantiga:
Hi Rick, fantastic! Unfortunately I don't think we can drop python 2.7 for the time being, but I'll be happy to take care about the merging once you think you're ready to roll out.
Apart from making python2 and 3 coexist, what are the things you see missing on the python 3 side?

rlizzo:
Thanks luca, I think at this point we have most of the basics covered. the things that are on the TODO list are: 

- [ ] make Tkinter both 2 & 3 compatable
- [ ] test which versions of python 3 the code supports (I'm aiming for fully compatability for python 3.3 and later versions, there were some issues in the 3.0, 3.1, and 3.2 release which don't play nice with later versions.)
- [ ] run tests to ensure that each operation in vmtkscripts works properly. 
- [ ] have CMAKE choose the correct version of python at compiletime 

Out of those four things, the only one I really need input from you is for the tests to run for vmtkscripts. I have been manual testing for the components I use in my day-to-day work: 

- imagereader/writer
- surfacereader/writer
- imagerenderer
- imagevoiselector
- imagesmoothing
- imagemorphology
- imageotsuthresholds
- levelsetsegmentation
- marchingcubes

But there's a ton of other modules that I've never actually needed to touch. In the absence of a test suite, I'm wondering if you have any sample code which takes advantage of the other library components that we could use as an informal test? This is going to be my main roadblock before giving it the all clear.

rlizzo:
Hey @lantiga , 

Thanks for your patience with my slow work on this. It's been a crazy few weeks. I have some time now in the next few weeks that I'll be using to polish everything up, but I wanted to give you a heads up that I'm close to being ready to roll out. 

Recently I've:
1) fixed some wires read/write behavior caused by modules not importing correctly

2) tested a fair bit of the code (still need to put together a formal test suite). I haven't so much focused on unittests as I have system test (making sure that each module works correctly). The PypeS module is fairly complex due to the way it passes around strings to dynamically execute code. Since ive always used dictionaries to handle similar situations, it's been quite a challenge to understand the flow and pathways code execution takes. It's a cool system though, and as I continue to get more familiar I'll be better able to write unittests for it. 

3) ive modified the top level CMakeLists.txt file to specify which python version to build against. To this end I needed to add some CONFIGURE_FILE options to specify the setup.py and vmtk.rb in the distribution module. I'd like for you to take a look at my last commit to make sure this is cool with you. I'm a novice at CMake so I may not have done this in the best way. 

I'm working on one last bug when using pypepad in python 2.7 environments. It's currently not parsing the input correctly, and I'm not really sure why. Hopefully I'll be able to make some headway tonight. 

The last bit worth mentioning is that in order to achieve full python 2-3 compatibility, i have used the "future" module extensively (ie. "from __future__ import xyz"). While this is technically part of the standard library, it is not active by default on most python 2.7 instillations. Typically you have to "pip install future" in order for it to work. Unfortunately I couldn't think of any reasonable way to get around this due to the significant changes in the way that imports and intiger division are handled in python 3. Hopefully this is ok for you.

Looking forward to your feedback.

rlizzo:
Well good news is that I fixed the word python2.7 pypepad issue. Had to do with the handling of Unicode literals. No problems there...

lantiga:
Hey @rlizzo, thank you for this final stretch of work! Really appreciated.
And sorry for my late reaction, I had a few things queued up. I'm reviewing tonight, hopefully we can merge very soon.

lantiga:
BTW, do you use conda? I was toying with the idea of moving to anaconda as the default package manager, so we have a uniform way to install on all platforms. This system Python + brew on OSX thing is breaking regularly one way or another. I'd be interested to hear your thoughts on this.

PypeS/pypescript.py
```diff
@@ -598,17 +607,25 @@ def IORead(self):
...
                     try:
-                        exec('import ' + member.MemberIO)
+                        writerModule = importlib.import_module('vmtk.'+member.MemberIO)
+                        # Find the principle class to instantiate the requested action defined inside the requested writerModule script.
+                        # Returns a single member list (containing the principle class name) which satisfies the following criteria:
+                        #   1) is a class defined within the script
+                        #   2) the class contains an attribute named 'self.SetScriptName'
+                        # This makes the assumption that only one class per vmtkscript file contains the self.SetScriptName attribute, and that
+                        # the class containing this attribute is the primary class to instantiate a new call to the scripts functionality
+                        writerModuleClasses = [x for x in dir(writerModule) if isclass(getattr(writerModule, x)) and hasattr(getattr(writerModule, x), 'SetScriptName')]
```
lantiga: Same as above. I'm fine with checking the method name too, I'd just like to know why you opted for this solution vs checking the parent class.

vmtkScripts/contrib/vmtkmeshwriter2.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+z#!/usr/bin/env python
```
lantiga: Looks like there's an extra `z` here

vmtkScripts/vmtkscripts.py
```diff
@@ -1,142 +1,144 @@
...
+    'vmtk.vmtkendpointextractor',
+    'vmtk.vmtkflowextensions',
+    'vmtk.vmtkicpregistration',
+    'vmtk.vmtkimagebinarize',
+    'vmtk.vmtkimagecast',
+    'vmtk.vmtkimagecompose',
+    'vmtk.vmtkimagecurvedmpr',
+    'vmtk.vmtkimagefeaturecorrection',
+    'vmtk.vmtkimagefeatures',
+    #'vmtk.vmtkimageinitialization',
```
lantiga: Few scripts in the list are commented out: what's the reason for it? Let me know if I can help if there's anything specific to be done about those scripts.

rlizzo: Hey luca, thanks for catching this! I thought I had taken care of it. 

There was an issue for a while with circular imports when running the scripts under Python 2. The six scripts that I had commented out each had an `from vmtk import vmtkscripts` statement at their header which was trying to execute when being imported in the `vmtkscripts` file - not great. I just rearranged the location of the import statement in these six sicripts, and everything seems to work well now for both python2 and 3.  Just pushed the change up now.

vmtkScripts/contrib/vmtkmeshwriter2.py
rlizzo: woops! Fixed.

PypeS/pype.py
```diff
@@ -236,20 +240,21 @@ def Execute(self):
...
-            except ImportError, e:
-                self.PrintError(e.message)
+                module = importlib.import_module('vmtk.'+scriptName)
+                # Find the principle class to instantiate the requested action defined inside the requested writerModule script.
+                # Returns a single member list (containing the principle class name) which satisfies the following criteria:
+                #   1) is a class defined within the script
+                #   2) the class contains an attribute named 'self.SetScriptName'
+                # This makes the assumption that only one class per vmtkscript file contains the self.SetScriptName attribute, and that
+                # the class containing this attribute is the primary class to instantiate a new call to the scripts functionality
+                scriptObjectClasses = [x for x in dir(module) if isclass(getattr(module, x)) and hasattr(getattr(module, x), 'SetScriptName')]
```
rlizzo: Hey Luca, 

I actually just pushed a fix for this to check for `issubclass(getattr(module, x), pypes.pypeScript)`. 

The reason I didn't do this in the first place was due to the same circular import error I was having  in the vmtkscripts module. This time it was the `pype.py` script importing 'pypes', which in turn imported `pype`... I didn't have a solution to that issue when I attempted to write this the first time, so just decided to check for `SetScriptName`. 

Now that i've got the imports all tightened up, the subclass solution works as expected.

rlizzo:
In response to your question about conda. Yes! It's all I use!

From a users perspective it's pretty much the perfect environment manager. Virtual env & pip can't even compare. It's stupid easy to use, and really takes all the work out of environment management. Probably the best part though is the fact that it can handle the binaries for native c/c++/fortran code. Seriously couldn't live without it at this point. 

I think my next project will be to try and set up a conda recipe. Should be interesting :)

lantiga:
I also started using conda recently, but I don't have experience in writing recipes. Let's kick it off as soon as we merge this. Which is... now :-)

lantiga:
Merged! Thank you so much Rick, we'll make the announcement today on the Google Group.

rlizzo:
Wow this was a great notice to wake up to! Thank you Luca! Glad to help!

lantiga:
FYI https://groups.google.com/forum/?fromgroups#!topic/vmtk-users/ecOhMaDyoQg

PR #179, opened 2017-05-30
Anaconda Pull

rlizzo:
This is a minor PR to merge changes made during prep for anaconda distribution. 

The most significant change is the full switch the the OpenGL2 rendering backend in the toplevent CMakeLists.txt file. This is required of VTK 7.1 on all platforms.


distribution/scripts/getbuildvmtk.py
```diff
@@ -22,7 +22,7 @@
 VTK_BUILD_DIR = os.path.join(WORK_DIR, 'vtk-build')
 ITK_BUILD_DIR = os.path.join(WORK_DIR, 'itk-build')
 VMTK_BUILD_DIR = os.path.join(WORK_DIR, 'vmtk-build')
-PARALLEL_JOBS = 3
+PARALLEL_JOBS = 7
```
lantiga: 7 sounds too much (I'm thinking about my laptop :-) ). I'd stick with 3 by default and eventually add a command-line flag to increase it.

rlizzo: good catch :) I adjusted it to the max cpu cores on the system - 1

lantiga:
Thank you!

PR #181, opened 2017-06-10
Numpy dataset adaptor

rlizzo:
Luca, 

This is a very early stage PR (not at all well tested or ready for release) to add increased integration to numpy arrays from vmtk. 

**Purpose:** To allow for the easy conversion of vtkDataObjects _to and from_ numpy arrays in a data-lossless manner (ie. preserving PointData/CellData array scalars and CellData connectivity). 

**Scope:** Conversion of surface data and centerlines data (eventually I'll get around to adding image data as well). 

**Methods:** 
1) Inclusion of 4 new vmtkscripts:
- `vmtksurfacetonumpy`
- `vmtknumpytosurface`
- `vmtkcenterlinestonumpy`
- `vmtknumpytocenterlines`

2) vmtk `Surface` and `Centerlines` data is accessed through the `vtk.numpy_integration dataset adaptor` method. `Points, PointData, & CellData` is copied into a series of numpy arrays contained within a structure defined by a nested python dictionary. 

3) a nested dictionary (of predefined structure) containing numpy arrays is converted into a `vtkPolyData` object in a method unique to the triangle cell connectivity of a `Surface` and a line cell connectivity of a `Centerline`

**Usage:** 
1) The script is intended to be run at the programmatic level (I haven't tried to save the output python dict object yet, but It's on the todo list). 

2) The following script can be used to run the conversion from vtkPolyData to the numpy array. 

_surface data_
```
numpyArrayConverter = vmtkscripts.vmtksurfacetonumpy()
numpyArrayConverter.Surface = foo.Surface
numpyArrayConverter.Execute()

arrayDict = numpyArrayConverter.ArrayDict
```
_centerline data_
```
numpyArrayConverter = vmtkscripts.vmtkcenterlinestonumpy()
numpyArrayConverter.Centerlines = foo.Centerlines
numpyArrayConverter.Execute()

arrayDict = numpyArrayConverter.ArrayDict
```
3) The dictionary output structure mimics the structure of the `vtk.numpy_integration datasetadaptor` output. The structure is:
```
arrayDict
    ['Points']                      <-- required. is Nx3 array of N vertexes and x, y, z locations
    ['PointData']                <-- required, even if subarrays are empty
        ['PointDataArray1']  <-- optional (ex. MaximumInscribedSphereRadius)
        ['PointDataArray2']  <-- optional
        ...
    ['CellData']                  <-- required
        ['CellPointIds']        <-- required. defines which cell conectivity to ['Points] Indices
        ['CellDataArray1']   <-- optional ie. CenterlineTractId
        ['CellDataArray2']   <-- optional
        ...
```
4) Any of the data in the dictionary can be modified, and new `PointDataArrays` or `CellDataArrays` can be added by simply specifying a new key value pair in the appropriate location in the dictionary. 

5) The dictionary can be converted into vtkPolyData format via calls to `vmtknumpytosurface` or `vmtknumpytocenterlines`, setting the parameter `InputDict = arrayDict`. The output data is a standard vtkPolyData format, and can be used as normal `Surface` or `Centerlines` data in following pypes. 

**Requirements:** 
1) Numpy must be installed on the users system. While this is an unsafe assumption for current vmtk distribution methods, this could easily be addressed in the anaconda package release by specifying numpy as a runtime requirement in the meta.yaml file. 

I wanted to get your input on this while it's in an early stage to see if you had any suggestions or concerns. These scripts have been a real help to me, and since this seems to be a popular topic on the vmtk-users forum I figured this would be useful to the community as well. 

Let me know.
-RIck


rlizzo:
Update:

A few bug fixes, but I also added two vmtkscripts to read and write the pickled ArrayDicts to a file. See the following files:
- `vmtknumpyreader`
- `vmtknumpywriter`

vmtkScripts/contrib/vmtknumpytocenterlines.py
```diff
@@ -0,0 +1,111 @@
...
+            points.InsertNextPoint(xyzPoint)
+
+        polyData.SetPoints(points)
+
+        self.PrintLog('converting numpy array to centerlines')
+        pointDataKeys = self.ArrayDict['PointData'].keys()
+        for key in pointDataKeys:
+
+            pointDataDType = str(self.ArrayDict['PointData'][key].dtype)
+            if pointDataDType.find('float') != -1:
```
lantiga: Here (and in all the other relevant spots) you could check the `dtype` in a more idiomatic way: https://stackoverflow.com/questions/22471644/how-do-i-check-if-a-numpy-dtype-is-integral

vmtkScripts/contrib/vmtknumpytocenterlines.py
```diff
@@ -0,0 +1,111 @@
...
+            else:
+                for pointData in self.ArrayDict['PointData'][key]:
+                    pointDataArray.InsertNextTuple(pointData)
+                polyData.GetPointData().SetActiveVectors(key)
+                polyData.GetPointData().SetVectors(pointDataArray)
+
+        cellDataArray = vtk.vtkCellArray()
+        numberOfCells = len(self.ArrayDict['CellData']['CellPointIds'].keys())
+        for cellId in range(numberOfCells):
+            numberOfCellPoints = self.ArrayDict['CellData']['CellPointIds'][str(cellId)].size
```
lantiga: I don't fully understand why you're required to use a string `cellId`. Does this come from the vtk util function that generates the dictionary?

vmtkScripts/contrib/vmtknumpytoimage.py
```diff
@@ -0,0 +1,88 @@
...
+from vmtk import vmtkrenderer
+from vmtk import pypes
+
+try:
+    import numpy as np
+except ImportError:
+    raise ImportError('Unable to Import vmtkimagetonumpy module, numpy is not installed')
+
+
+# convenience class for nested disctionaries
```
lantiga: typo: "disctionaries"

vmtkScripts/contrib/vmtknumpytoimage.py
```diff
@@ -0,0 +1,88 @@
...
+                pointDataArray = vtk.vtkFloatArray()
+            if pointDataDType.find('int') != -1:
+                pointDataArray = vtk.vtkIntArray()
+
+            pointDataArray.SetNumberOfComponents(1)
+            pointDataArray.SetName(key)
+
+            flatArray = self.ArrayDict['PointData'][key].ravel(order='F')
+            for pointData in flatArray:
+                pointDataArray.InsertNextValue(pointData)
```
lantiga: If you know how many elements you'll add to `pointDataArray`, it's faster to pre-allocate memory with `pointDataArray.SetNumberOfValues(...)` and the add the values with `pointDataArray.SetValue(i,pointData)`.
`InsertNextValue` is slower because it will likely lead to re-allocations and memcpy's as it grows.

lantiga:
Hi Rick, sorry for the late feedback, busy time.

I fully support this addition, I've come across quite a few users who would have liked to mess with data, but they were not into delving into VTK data structures.

As to writing/reading, you can just `np.save` and `np.load` the dictionary of arrays. We could eventually also include hdf5 (not as a dependency, just as an output option if the module is available), but npy would suffice for now.

vmtkScripts/contrib/vmtknumpytocenterlines.py
rlizzo: Excellent suggestion! Fixed for:
- `vmtknumpytocenterlines`
- `vmtknumpytoimage`
- `vmtknumpytosurface`

in the latest commit i'm about to push.

vmtkScripts/contrib/vmtknumpytocenterlines.py
rlizzo: This is a result of how I've chosen to store the cell data in the nested dictionary. 

Each cell in the vtkDataObject is written as a dictionary key into `self.ArrayDict['CellData']['CellPointIds']`. Each cell is identified by a zero-indexed intiger, so for a centerline with 3 cells, there will be arrays stored in each of the following dictionary keys:
- `self.ArrayDict['CellData']['CellPointIds']]'0']`
- `self.ArrayDict['CellData']['CellPointIds']]'1']`
- `self.ArrayDict['CellData']['CellPointIds']]'2']`

Since the loop variable `cellId` in `for cellId in range(numberOfCells):` is an intiger type, we need to convert it to a string type (ie. `self.ArrayDict['CellData']['CellPointIds'][str(cellId)]`) in order to access the array stored in that dictionary key. 

Does that help answer your question?

vmtkScripts/contrib/vmtknumpytoimage.py
rlizzo: Thanks! Fixed.

vmtkScripts/contrib/vmtknumpytoimage.py
rlizzo: Great Suggestion Just fixed this as well.

rlizzo:
Thanks luca, 

I considered using `np.save` and `np.load`, but I chose not to use them since it does not save the same python object that `pickle` does. When saving an `npy` file, it converts the `ArrayDict` "Dictionary of Arrays" to an "Array of Dictionaries of Arrays". This is an incompatible change to the read/convert methods, and I didn't think there was any point to it since:

1)  the `pickle` module is part of the standard library, and people should be used to using it
2) There would be no space saving, since we are specifying a binary file type in the `wb` pickle flag. 

Let me know if/why you might disagree? Would be interested to learn why you might prefer npy.  

HDF5 is a great idea as well. There's a lot we could do with that to store the entire contents of the analysis pypeline from initial image, any filtering performed, surface extraction, and centerline/mesh generation. Would be cool to use the format to store the entire anslysis history the was performed.

vmtkScripts/contrib/vmtknumpytocenterlines.py
lantiga: Hi Rick, I would suggest not to use a dictionary after 'CellPointIds', but a simple list, that you can index with an integer. You could use an integer as a key to the dictionary also, but there's no reason to do that, since `cellIds` are sequential and start from `0`. In any case, the `str(cellId)` doesn't really look good and I'd avoid it.

On a more general note, this layout (one entry per cell) is going to use more memory than needed for large datasets with many cells. It's not going to make a big difference for centerlines, but as a general strategy for other kinds of datasets it's going to be inefficient.

You could store cells more efficiently by using one or two arrays.
One-array solution: store cells sequentially in a NumPy array like 
```
cell0npts cell0pt0 cell0pt1 ... cell0pt4 cell1npts cell1pt0 cell1pt1 ...
```
This is quite compact, but it gets tricky if you every need to access a cell given its id.

Two-array solution: the above, plus an array with size `ncells` containing the index of each cell in the above array.

Last note: if you mix cell types (lines, polygons, etc) you might want to also have an array of size `ncells` containing the cell type for each cell.

It's your call, my only request is to get rid of the cellId as string thing. Thanks!

lantiga:
Great, both Pickle and HDF5 👍

vmtkScripts/contrib/vmtknumpytocenterlines.py
rlizzo: Thanks for the feedback, That makes much more sense. 

My other thought is to scrap dictionaries entirely and use a structured numpy array to store all the data - I'm experimenting with this now in a jupyter notebook I'll share with ya, but I think this would be a good solution as well since:
1) Access would be essentially the same due to named fields
2) Numpy array storage is about as efficient as you can get without going into the Cython realm

I was planning on formalizing this approach tonight, and will update you at the end. May be totally rewriting most of these scripts depending on the performance. 

Look for an invite to a private repo i've created to host all my experimentation work.

rlizzo:
Hey luca, 

I've been using this module on my own system fairly consistently for the last few weeks. Wondering if you've had a chance to give it a shot since I made the changes you requested?

Also added a tutorial as a PR at https://github.com/vmtk/vmtk.github.com/pull/4 so that this doesn't go undocumented.

lantiga:
Reviewed and merged! Thank you Rick!

rlizzo:
Awesome! Thanks luca!

Issue #183, opened 2017-06-25
Improper string comparison for vmtksurfacesmoothing.py


normanius:
**Issue**: In certain circumstances, vmtksurfacesmoothing fails with

> Error: smoothing method not supported.

although the string containing the method is correct: `vmtkSurfaceSmoothing.Method == 'taubin' ` yields `True`.

**Details**: In [vmtksurfacesmoothing.py, around line 63](https://github.com/vmtk/vmtk/blob/master/vmtkScripts/vmtksurfacesmoothing.py#L63), the method string is compared with the wrong string comparison. The issue is described nicely in [this stackoverflow article](https://stackoverflow.com/questions/1504717). Indeed,`vmtkSurfaceSmoothing.Method is 'taubin' ` yields `False` in my case (where the method string is read as a command-line argument using argparse). 

**Solution**: 
Change `if self.Method is '<method>'` to `if self.Method == '<method>'`, with `<method>` being either `'laplace'` or `'taubin'`. This is considered the proper way of comparing strings literally in python.

rlizzo:
Thank you @normanius. I just created pull request #187 to address this issue.

PR #187, opened 2017-07-03
fixed improper string comparisons referenced in vmtk/vmtk issue #183

rlizzo:
Fixed improper string comparison referenced in Issue #183 

Also fixed bug in vmtkpathlineanimator.py

Thank you to @normanius for identifying the problem and proposing a solution

@lantiga please merge at your convenience.


lantiga:
Whoops, thanks!

PR #188, opened 2017-07-05
Compatibility with VTK 8.0

deparis:
I am trying to build vmtk with brew on MacOS X.
I have found two issues:

./distribution/homebrew/vmtk.rb :
- is not pointing to the last release
- does not recognize python 

VTK_MAJOR_VERSION : brew installs VTK 8.0.0 but the MACROS for 7 and later are set for MINOR => 1

I have fixed the second issue. For the first one I only have a hard fix on the file.


lantiga:
Thank you @deparis!

PR #189, opened 2017-07-22
removed looping from numpy image conversion. significant speedups.

rlizzo:
fix in reference to https://github.com/vmtk/vmtk.github.com/pull/4#issuecomment-317213285 . 

by using the numpy dataset adaptor module we are getting > 2 orders of magnitude speedup on the conversion.


rlizzo:
Tested and merged in response to prior conversation.

PR #196, opened 2017-10-09
New script: vmtksurfaceclippercenterline

kayarre:
adapted the vmtksurfaceendclipper to use centerline geometry information to get a more accurate normal to clip a surface at the inlets and outlets.

### DIFF
--- a/vmtkScripts/CMakeLists.txt
+++ b/vmtkScripts/CMakeLists.txt
@@ -116,6 +116,7 @@ SET(SCRIPTS_SRCS
   vmtksurfacedecimation.py
   vmtksurfacedistance.py
   vmtksurfaceendclipper.py
+  vmtksurfaceclippercenterline.py
   vmtksurfacekiteremoval.py
   vmtksurfaceloopextraction.py
   vmtksurfacemassproperties.py
@@ -187,4 +188,3 @@ INSTALL(PROGRAMS ${STRIPPED_SCRIPTS_SRCS}
   COMPONENT RuntimeExecutables)
 
 ENDIF (NOT WIN32 AND NOT VMTK_MINIMAL_INSTALL)
-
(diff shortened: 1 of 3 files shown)


rlizzo:
Hi @kayarre, sorry for the delayed review. 

A few minor issues on first glance: 
1) contributed scripts should be placed in the `/vmtk/vmtkScripts/contrib` directory instead of the `/vmtk/vmtkScripts/` directory. 

2) There are quite a few sections of code in the `vmtkSurfaceClipperCenterline.Execute()` method which are just bulk commented out from your adaptation of the `vmtksurfaceendclipper.py` script. If these are left intentionally commented out can you add comments explaining why? Otherwise just please delete. 

3) In [the following lines](https://github.com/kayarre/vmtk/blob/f45a8fd38b071bbce0f6f696c5c2819e16f0a22b/vmtkScripts/vmtksurfaceclippercenterline.py#L264-L266). 
```
        for seed_type in seeds_dict.keys():
            for i in range(seeds_dict[seed_type].GetNumberOfIds()):
                seedId = seeds_dict[seed_type].GetId(i)
```
This is not particuarly easy code to read (especially since you are storing a vtk object as a value in a python dict. First suggestion: please add a comment for future readability about what you are doing. 
Second, to keep the code both python 2 and 3 compliant (along with increasing readability) I would recommend the following change:
``` 
    for (seed_type, target_seed_ids) in seed_dict.items():
        for i in range(target_seed_ids):
            seedId = target_seed_ids.GetId(i)
            ...
```
That is a bit more pythonic, and makes sure that the differences in the way python 2 and 3 handle the dictionary keys() method is not an issue. 

Also, can you provide an example or two where this method yeilds more accurate results than the current `vmtksurfaceendclipper.py` script? If that is the case, it might just be worth integrating it into the main clipper script (potentially as an option), rather than keeping two of essentially the same script around.

Thanks for the contribution! It is much appreciated!
Rick

kayarre:
Thanks @rlizzo for the feedback. I have created another pull request to update the vmtksurfaceendclipper to include this functionality and address your comments.

The normal estimation routine is poor and the clipping direction is often inverted.

I am going to close this request.

PR #199, opened 2017-10-17
comment and small initialization bug

kayarre:
added comment and small fix to initialization bug.


rlizzo:
Thanks for the bugfix @kayarre! This looks good to me - merged!

PR #200, opened 2017-10-18
Modify vmtksurfaceendclipper to clip based on centerline information

kayarre:
per comments from @rlizzo in #196 I added clipping using centerline information feature to the  vmtksurfaceendclipper.

This modification provides a means to use the normal information from the frenet frame of reference along the centerline to clip at inlet and outlet locations. The seed selection requires that the users knows the inlet and outlet direction used in the centerline generation such that the proper normal direction is applied in the clipping.

The issue with the current implementation of surfaceendclipper is two things.
1. The normal estimation is poor
2. The normal may be inverted and clip the whole surface. The vtkClipPolyData section could be updated to catch the instance when the normal is inverted and select the proper clipped body, however It seemed easier to use the centerline information.

The additions also includes removing the use of python dictionaries in the  for loop.


rlizzo:
Thanks for this PR @kayarre! I'm really sorry it took so long to review this! It works really well. Nice Job!

Also, I meant to just push a few small changes to your PR before merging it (just updated the docstring and simplified some expressions), but it appears that I have royally screwed up and somehow merged in every commit to master from the point you branched to now...  Not entirely sure how that happened. Sorry about the confusion, will fix now.

rlizzo:
Ok. history crisis averted by rebasing on master. Merging now!

PR #205, opened 2017-12-06
removed problem python wrapping modules from vtk wrapping

rlizzo:
test for slicer extension fix.


lassoan:
Thank you.

It should work, but I'll confirm tomorrow morning when the nightly dashboard build results are available.

jcfr:
Can this be merged ? 

@lassoan  This is what is causing build error and prevent the associated extension from building.

lassoan:
Yes, it works well on Windows and at least it's built correctly on Linux. (there is a Mac build error due to some recent changes in Slicer, but we'll fix that in Slicer)

rlizzo:
Thanks @lassoan. Merged

PR #207, opened 2017-12-13
Update cmake policies 

rlizzo:
This change is in relation to a PR about to be submitted to the vmtk/conda-recipes channel. 

Minimum version was updated, and CMake Policy CMP0025 is set to NEW in
order to allow us to detect the features of custom compiler packages.
This is a requirement for using the anaconda c and cxx compilers, as
they implement a spec-compliant, but non standard compiler set on OSX
and Linux in order to standardize the compilers used to build non python
dependencies.

Also needed to set CMP0033 to OLD behavior. This allows for us to
continue using the cmake command export_library_depends(), which was
replaced in cmake 3.0+. Failure to set this behavior yields a fatal
error.

Requesting review from @lantiga to verify that no unexpected side effects will affect the calculations as a results. In addition, @lassoan, can you please verify that this does not break the slicer build?


lassoan:
Thank you, I'll try this and let you know by tomorrow morning if everything works well.

lassoan:
Build is OK on Windows, Linux, and Mac. Packaging is OK on Windows and Linux.

Packaging is still broken on Mac (http://slicer.cdash.org/viewBuildError.php?type=0&buildid=1153075) but that problem is not related to these changes but cause by recent changes in Slicer and should be fixed there. @jcfr Could you please have a look at the nightly VMTK extension packaging on Mac?

rlizzo:
Thanks @lassoan . Are you compiling against python 3 libraries by chance? Because I've also been running into the strangest packaging issues for macs.  

When we build against 2.7, it works perfectly, but when we compile against python 3.5 or 3.6, the build completes, everything gets installed correctly. However, on runtime the interpreter encounters a fatal `pythreadstate_get()` fault - usually an indication that you are using an interpreter version you didn't compile against.

 Here's were it gets weird.  when I set up my virtual environment to build against, I'll specify py3.6, and it will install 3.6.3 (or whatever the latest sub-release is). I'll use this to compile, and if I run from that same environment I get the aforementioned result. However if I then manually switch out python interpreters and downgrade to python 3.6.0 or 3.6.1, it works perfectly. (same thing for py3.5). Even stranger is if I do the same steps, don't downgrade, but pull up a `vtkpython` interpreter from the command line (which will be the same as the build interpreter, aka 3.6.3), everything works perfectly. 

So while we have a solution by pinning 3.6.0 or 3.6.1 as a hard runtime requirement, it's a total bodge of a fix that's just rather annoying. Does this sound like anything you guys are seeing?

lassoan:
This sounds very odd. We still use Python 2.7 (and will not switch for at least a few more months) and the error that we have is different.

rlizzo:
Yeah I've been tearing my hair out over this, and really don't understand much more than I just shared. It's not even clear where the issue is being introduced, is it VMTK, CMake, VTK, Conda, Python or some strange combination of them? 

 Just thought I'd ask. Thanks!

jcfr:
If you would like help improving your build system and have install-able python wheels (e.g pip install vmtk), and  conda package. Let me know . At Kitware, we have been developing  http://scikit-build.org/ for some time and are successfully using it to create ITK wheels (https://blog.kitware.com/itk-python-wheels/) and soon VTK wheels.

rlizzo:
@jcfr we have conda packages, but our python wheels are for the last release. I'll see if I can get the wheels built and will give ya a shout. Looks like an interesting project. Thanks!

jcfr:
>  export_library_depends()

I suggest to update the build system to not use this deprecated function.

jcfr:
This PR can be closed. It is superseded by https://github.com/vmtk/vmtk/pull/209

rlizzo:
Closed.

PR #211, opened 2017-12-20
Modernize cmake export commands.

rlizzo:
Removed dependency on setting CMP0022 and CMP0033 to the old behavior
(which is giving deprication notices on cmake 3.9+). To do this, the
cmake export_library_dependencies needed to be removed. Instead of
keeping this at the top level cmake lists file, it is replaced by an
install export command, which pulls from VMTK-Targets. The VMTK-Targets
exports are defined from install commands which process c++ targets.

Build is tested and working on Mac and Linux.


rlizzo:
@lassoan if you could test this as well it would be appreciated.  This should be the last of the build system modifications we need to make for awhile. Sorry to keep pulling you in on this!

lassoan:
Not a problem at all. I'll test it later tonight and let you know ow how it works.

lassoan:
I've tested this on Windows and it all works well. We are transitioning from VTK7 to VTK9 on our nightly dashboard, so it'll take a couple of days to do further testing on other platforms.

rlizzo:
Thanks @lassoan. 

@lantiga requesting review. I have tested building on Linux and Mac via command line cmake and with https://github.com/vmtk/conda-recipes/pull/6. Everything seems to be in order, but want to make sure that you are ok with the changes before merging something which modifies how external projects will include the vmtk library files (even if CMake should handle the details for us).

PR #218, opened 2018-01-10
[WIP] Testing and CI!

rlizzo:
This is an early stage PR which starts working on a basic testing method for the vmtkScripts and Pypes module. Py.Test is used as the testing framework. Documentation, explanation of methods, and (many more) test additions to follow.


rlizzo:
Before I go further, I'd like your input on my approach @lantiga &/or @daron1337. 

**Run Instructions**
create a conda environment with the latest version of vmtk pushed to anaconda cloud. activate the environment, and install pytest.

checkout the PR in a local branch, and from the root of the repo run the following command `py.test ./tests/`. If you want to view the `stdout` content, append the `-s` flag to the command. 

at the time of writing, you should see 82 tests pass and 1 test xfail. 

**Methods**
I'm going to focus primarily on the python vmtkscripts modules for the time being - after we build up the test base I'll move onto the pypes modules. 

The data I am using is publicly available as part of the VTK Testing Dataset. It can be acquired via ` git clone git://vtk.org/VTKData.git VTKData`. 

In order to make the tests run as fast as possible I only am only storing the input data files in the actual repository. Rather than keep a large amount of "processed" (ie result) files around (which we would have to read in and compare to the processed results at the end of every test going), at the end of every test, I take the processed VTK object, convert it to an equivalent numpy ndarray, and taking a hash of the array. This way, we only need to run a simple 

```python
assert {processed_hash} == {known_hash}
```
to know if we got the right result for the test. 

I am also making heavy use of pytest fixtures and test parameterization. The parameters are defined as decorators on the test functions, the fixtures are stored in the `conftest.py` file. 

While this seems to be a fairly efficient solution, it may not be the easiest test setup to understand. However, I'm not sure that any solution would be particularly pythonic (having to deal with vtk objects pretty much guarantees that).  Wanted to get your thoughts before moving forward down this path.

rlizzo:
[Testing](https://trello.com/c/SGQZT1Of/1-testing)

lantiga:
Hi @rlizzo, great job! I like the approach of keeping the repository lean.

Having test data in the main repository is ok as long as we keep it small, so it's good for now.

Just so you know, the way ITK does it (see https://github.com/InsightSoftwareConsortium/ITK) it keeps test data in a separate location and it uses CMake to pull the external data in the `.ExternalData` repository: 
see
https://github.com/InsightSoftwareConsortium/ITK/blob/ba79114d4fb56ad065a68a09db4463f8af852f7e/CMake/ExternalData.cmake
and
https://github.com/InsightSoftwareConsortium/ITK/blob/5165c587670fadb679a35730b4902af852166e69/CMake/ITKExternalData.cmake.

Another way would just be to keep a separate repo with the data and pull it automatically when tests are run. It's not essential for now in my opinion.

As for comparing hashes, I like that approach. However, I think we need to store those baselines somewhere, just to understand how the output was once we see that a test has failed. A separate repo would be good in this regard. What do you think?

rlizzo:
Thanks for the feedback @lantiga! I figured out pretty quickly why just comparing hashes doesn't seem to be utilized in more packages. 

While the tests work consistently within one version of VTK/ITK & operating system, subtle differences in the algorithm/operating system can result in small differences in the last few decimal places of the image data, and any difference (no matter how small) results in a completely different hash, and the tests failing when they really shouldn't. 

I'm transitioning to a similar method as the ITK repo, where the reference images are stored in an external project. I'm transitioning from a hash based system to an image comparison based system now. Since I don't have sufficient permissions to create a new repo for the vmtk organization, the data files are stored under my github account at https://github.com/rlizzo/vmtk-test-data. 

Please clone that repo and set it up under the github vmtk organization (also please attach a license file to it) I'll take care of filling out the readme and documenting how to test as the structure of the repo matures.

tests/test_vmtkScripts/conftest.py
```diff
@@ -0,0 +1,301 @@
...
+from hashlib import sha1
+import copy
+
+
+@pytest.fixture(scope='function')
+def input_datadir():
+    '''
+    returns a path to the vmtk/tests/testData directory
+    '''
+    datadir = '/Users/rick/projects/vmtk/vmtk-test-data/input'
```
rlizzo: Until I get the CMake external project set up, the paths to the test-data is hardcoded to my system. If you want to run these test locally, please change these to reflect the appropriate values on your system.

tests/test_vmtkScripts/conftest.py
```diff
@@ -0,0 +1,301 @@
...
+        converter.Execute()
+        return converter.ArrayDict
+    return make_poly_to_np
+
+@pytest.fixture()
+def write_image():
+    def make_write_image(image, filename):
+        writer = imagewriter.vmtkImageWriter()
+        writer.Image = image
+        writer.OutputFileName = os.path.join('/Users/rick/projects/vmtk/vmtk-test-data/imagereference', filename)
```
rlizzo: Another hardcoded path is here

tests/test_vmtkScripts/conftest.py
```diff
@@ -0,0 +1,301 @@
...
+        writer.Execute()
+        return
+    return make_write_image
+
+
+@pytest.fixture()
+def compare_images():
+    def make_compare_images(image, reference_file):
+        reader = imagereader.vmtkImageReader()
+        reader.InputFileName = os.path.join('/Users/rick/projects/vmtk/vmtk-test-data/imagereference', reference_file)
```
rlizzo: another hardcoded path is here

tests/test_vmtkScripts/conftest.py
```diff
@@ -0,0 +1,301 @@
...
+        reader.Execute()
+
+        comp = compare.vmtkImageCompare()
+        comp.Image = image
+        comp.ReferenceImage = reader.Image
+        comp.Method = 'subtraction'
+        comp.Tolerance = 0.1
```
rlizzo: TODO Note: Allow for configuring the tolerance on a "per-test" basis

rlizzo:
@lantiga, I've gone ahead and gotten the external data project working for building the tests. I followed the ExternalData process from ITK almost exactaly. Data is stored in the main repository as a SHA512 hash of the binary object; the `filename.extension.sha512` files are generated with the new [hash utilities](https://github.com/rlizzo/vmtk/blob/d09c0b89aabf198487c773460ae639c124589e34/tests/utilities/hashdata.py) command line program. 

Input data is taken from the https://github.com/rlizzo/vmtk-test-data repo, and a hash is stored in the `vmtk/tests/data` folder. The original input data files are then uploaded to [kitwares datastore]( https://data.kitware.com/#collection/5a62be7f8d777f0649e028e6) under the newly created VMTK collection (register for this and send me your username so I can add you as an admin). We call into the [datastore API](https://data.kitware.com/api/v1#!/file/file_downloadWithHash) with the hash from each file in `tests/data/*` during the build process, and download the file to the `CMAKE_BINARY_DIR/ExternalData`. Tests are copied from the VMTK source dir to the `CMAKE_BINARY_DIR/tests` folder, and the appropriate paths to data files are configured. 

This seems to work fairly well so far, but I'm looking forward to hearing your feedback.

rlizzo:
Hey @lantiga. Once these builds pass I think I'm ready for you to give it a final review and merge. If you could just take a look and make sure I'm not doing anything totally insane somewhere, I'd appreciate it. 

I've activated travis CI and appveyor for the VMTK repo, so when this is merged to master, it should build and populate the build passing status so we have a baseline for the next set of contributions. Let me know if there's any issues. 

Thanks!

rlizzo:
@lassoan, can you test this with the slicer build?

rlizzo:
FYI, I just rebased onto master to incorporate the last few PR changes.

rlizzo:
FYI @lantiga, after our discussion today, I am performing this merge now.

lantiga:
Fantastic job @rlizzo !

lantiga:
Oh, BTW, are you on Twitter? I'd like to start pushing news out.

PR #231, opened 2018-02-07
scaling text for high dpi displays

rlizzo:
Needs testing on linux/windows with standard 1080p displays. works on retina macbook.


rlizzo:
This appears working to me. Merging.

Issue #229, opened 2018-02-07
Error in branchclipper


rlizzo:
Placeholder to remind me to fix this. 

```
_____________________ test_clip_groupids_with_insideout_surface_returns_as_expected[groupids0-0] _____________________

aorta_centerline_branches = (vtkCommonDataModelPython.vtkPolyData)0x11cae9ca8
aorta_surface = (vtkCommonDataModelPython.vtkPolyData)0x11cae9e88, groupids = [0], paramid = '0'
compare_surfaces = <function compare_surfaces.<locals>.make_compare_surface at 0x11cad3950>
write_surface = <function write_surface.<locals>.make_write_surface at 0x11caee950>

    @pytest.mark.parametrize("groupids,paramid", [
        ([0], '0'),
        ([2], '2'),
        ([3], '3'),
        ([0, 2], '02'),
        ([2, 3], '23'),
    ])
    def test_clip_groupids_with_insideout_surface_returns_as_expected(aorta_centerline_branches, aorta_surface,
                                                                     groupids, paramid, compare_surfaces, write_surface):
        name = __name__ + '_test_clip_groupids_' + paramid + '_with_insideout_surface_returns_as_expected.vtp'
        clipper = branchclipper.vmtkBranchClipper()
        clipper.Centerlines = aorta_centerline_branches
        clipper.Surface = aorta_surface
        clipper.RadiusArrayName = 'MaximumInscribedSphereRadius'
        clipper.BlankingArrayName = 'Blanking'
        clipper.GroupIdsArrayName = 'GroupIds'
        clipper.GroupIds = groupids
        clipper.InsideOut = 1
>       clipper.Execute()

test_vmtkScripts/test_vmtkbranchclipper.py:78:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <vmtk.vmtkbranchclipper.vmtkBranchClipper object at 0x11cb0d7f0>

    def Execute(self):

        if not self.Surface:
            self.PrintError('Error: No input surface.')

        if not self.Centerlines:
            self.PrintError('Error: No input centerlines.')

        if self.Interactive and not self.vmtkRenderer:
            self.vmtkRenderer = vmtkrenderer.vmtkRenderer()
[...]

PR #233, opened 2018-02-13
fix branchclipper with insideout enabled. 

rlizzo:
there was no method named apropriatly before.


rlizzo:
fixes https://github.com/vmtk/vmtk/issues/229

PR #238, opened 2018-02-19
no longer need to modify ld_library_path or dyld_library_path

rlizzo:
this ensures that the required libraries are always found independent
from DYLD_LIBRARY_PATH and the install location for local MacOS builds.
CMake is able to specify the path were vtkvmtk*.so executables should
look for their associated shared libraries. This is a cleaner way to
make the library work for local development, and is a better way to work
with conda environments (as anaconda does some odd things to remap the
default shared library load paths).

This works with both a full build from source, or a bootstrapped build
with vtk/itk preinstalled/used from a conda environment


rlizzo:
@lantiga, merging this since this now works. thanks for the debugging help earlier!

lantiga:
Nice 🥇

PR #239, opened 2018-02-19
[WIP] CCache on ci

rlizzo:
This is a test to see if CI builds can be speed up by using ccache. do not pull from this branch. there will be many changes and likely a few rebases.


rlizzo:
Forgoing this for now. Does not seem like a good idea with conda-build

PR #240, opened 2018-02-20
Test data asset management via submodules

rlizzo:
### Test data asset management via submodules  

This change is designed to simplify the management of binary test data assets. Previously, we used a system based on cmake externalData downloads which required the manual management of file versions, hashes, and download servers. In this model, we simply use a git submodule with the binary data stored in it. If the testing flags are enabled, the submodule is cloned at build time, and the test data paths are automatically configured to reflect the correct install location. Tests can be run locally, or via CI services.


rlizzo:
@lantiga, this build passes, and I believe it is a drop in replacement for the External Data Module. I'm ready to merge, but would like to make sure you agree with the structure and documentation of the process before merging.

PR #253, opened 2018-03-23
add invert to transform surface from RAS back to XYZ 

kayarre:
There is a use case where Slicer registration tools are used to map different image modalities to the same image space. however if the segmentation is performed on the un-transformed image there is not an easy mapping between the different coordinate frames that such that paraview can be used.

adding invert to vmtksurfacetransformtoras.py provides the following workflow.
1. extract surface from a image volume (DSA for example) using VMTK
2. transform surface to ras (normal use case) using VMTK
3. manually/automated registration in RAS space (slicer) of DSA to MRI (SLICER)
4. transform surface in RAS space using registration transforms (SLICER)
5. apply inverse xyz2ras transform from the registered image to the registered surface to get surface that matches space of image in paraview (VMTK)

caveat: The registered image volume must be converted to the vti format with vmtk. Paraview can't read compressed nrrd format (slicer default) and mha format doesn't match space (Paraview issue of vmtkimagewriter may have bug see #254 )

### DIFF
--- a/vmtkScripts/vmtksurfacetransformtoras.py
+++ b/vmtkScripts/vmtksurfacetransformtoras.py
@@ -6,7 +6,7 @@
 ## Date:      $Date: Sun Feb 21 17:02:37 CET 2010$
 ## Version:   $Revision: 1.0 $
 
-##   Copyright (c) Luca Antiga, David Steinman. All rights reserved.
+##   Copyright (c) Luca Antiga. All rights reserved.
 ##   See LICENCE file for details.
 
 ##      This software is distributed WITHOUT ANY WARRANTY; without even 
@@ -28,12 +28,14 @@ class vmtkSurfaceTransformToRAS(pypes.pypeScript):
 
         self.Surface = None
         self.XyzToRasMatrixCoefficients = None
+        self.InvertMatrix = 0
 
         self.SetScriptName('vmtksurfacetransformtoras')
         self.SetScriptDoc('transform a surface generated in XYZ image space into RAS space')
         self.SetInputMembers([
             ['Surface','i','vtkPolyData',1,'','the input surface','vmtksurfacereader'],
-            ['XyzToRasMatrixCoefficients','matrix','float',16,'','coefficients of XYZToRAS transform matrix']
+            ['XyzToRasMatrixCoefficients','matrix','float',16,'','coefficients of XYZToRAS transform matrix'],
+            ['InvertMatrix','invert','bool',1,'','invert matrix before applying transformation']
             ])
         self.SetOutputMembers([
             ['Surface','o','vtkPolyData',1,'','the output surface','vmtksurfacewriter']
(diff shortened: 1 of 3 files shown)


lassoan:
ParaView is a great software, but it is barely usable for medical image computing (lack of image orientation, DICOM, compressed mhd/nrrd support, sub-par volume rendering, lack of non-linear transform support, no segmentation and registration tools, etc). 3D Slicer has none of these limitations. I would recommend you to give it a try instead of struggling to use ParaView for tasks that it is not designed to do.

kayarre:
@lassoan thanks for the helpful feedback. 3D Slicer has been immensely helpful. I think the challenge I am facing is that I need a reasonably good visualization tool (e.g. paraview) that I can simultaneously view CFD results on top of the image volume. So I agree with you, I am doing most things in Slicer.
 
This use case came about because I don't want to have to re-segment images that were segmented pre-registration.

lassoan:
You can visualize vector fields very nicely in Slicer (see https://www.slicer.org/wiki/Documentation/Nightly/Modules/Transforms#Transform_display_options), you can visualize volumetric meshes, and time sequences of volumes, models, vector fields. You also have access to all VTK and functions in Slicer, so you can set up streamlines or other visualizations by a couple of lines of Python code. If you have any questions about these, write to https://discourse.slicer.org.

vmtkScripts/vmtksurfacetransformtoras.py
```diff
@@ -28,12 +28,14 @@ def __init__(self):
...
         self.XyzToRasMatrixCoefficients = None
+        self.InvertMatrix = 0
 
         self.SetScriptName('vmtksurfacetransformtoras')
         self.SetScriptDoc('transform a surface generated in XYZ image space into RAS space')
         self.SetInputMembers([
             ['Surface','i','vtkPolyData',1,'','the input surface','vmtksurfacereader'],
-            ['XyzToRasMatrixCoefficients','matrix','float',16,'','coefficients of XYZToRAS transform matrix']
+            ['XyzToRasMatrixCoefficients','matrix','float',16,'','coefficients of XYZToRAS transform matrix'],
+            ['InvertMatrix','invert','bool',1,'','invert matrix before applying transformation']
```
rlizzo: Please add a better description of why this option would be used.

vmtkScripts/vmtksurfacetransformtoras.py
```diff
@@ -6,7 +6,7 @@
 ## Date:      $Date: Sun Feb 21 17:02:37 CET 2010$
 ## Version:   $Revision: 1.0 $
 
-##   Copyright (c) Luca Antiga, David Steinman. All rights reserved.
+##   Copyright (c) Luca Antiga. All rights reserved.
```
rlizzo: Don't modify the copyright owners.

rlizzo:
Hey kurt, these changes look good to me, just two minor things which need to be addressed before we merge!

rlizzo:
Linking this to https://github.com/vmtk/vmtk/pull/255. Planning to merge both at the same time.

rlizzo:
Hey Kurt. Just wanted to let you know I added a few tests. Am merging this now! Thanks for the hard work!

kayarre:
Thank you so much Rick I have been swamped the past couple weeks.
-Kurt

Issue #254, opened 2018-03-23
vmtkimagewriter not consistent between formats


kayarre:
writing the same image from nrrd to vti or nrrd to mha each have different origins.

from vti: Origin="32.695867086 41.89599113 2.6593247071"
from mha: Offset = -32.695867085974825 -41.895991130092369 2.6593247071163644

Things subsequently don't line up in paraview using the mha format, but using the vti format things line up correctly.

example commands:
vmtkimagewriter -ifile DSA_trans_final.nrrd -ofile DSA_trans_final_vmtk.vti
vmtkimagewriter -ifile DSA_trans_final.nrrd -ofile DSA_trans_final_vmtk.mha

rlizzo:
Thanks for pointing this out Kurt; I think I need to play around with it a bit to understand the issue. Could you upload an example data file to this dropbox folder?

https://www.dropbox.com/request/XQBg7fPjBFeFgViQpb0g

Before I dig in, I'm wondering if your issue can be solved with the `ApplyTransform` and `RasToIjkMatrixCoefficients` options in vmtkImageWriter?  (I think this PR may be related? https://github.com/vmtk/vmtk/pull/155)

lassoan:
Nowadays all medical image computing software pretty much standardized on using LPS coordinate system. Even software that use RAS internally (such as 3D Slicer) uses LPS coordinate system in files, and converts to internal coordinate system on import.

What software do you use that would expect data in RAS coordinate system?

kayarre:
@rlizzo  I uploaded and image volume and will experiment with what you suggest, many thanks. 

@lassoan I didn't realize that the internal vs read/write were different. I think the problem is that slicer handles the images properly but it doesn't do the same thing for say a surface mesh. When a surface mesh is imported into slicer there is no conversion to RAS therefore it mush be pre-transformed and when you write the surface after slicer transformation it's written as is (as far as I can tell) it's still in RAS. So I need a way for things to end up in the same space when I export them and view them (Paraview in my case) Thank you for the feedback.

kayarre:
@rlizzo using the ApplyTransform option only works for VTKIO formats (nrrd, mha... etc)
I have implemented a work around by adding vtkXMLStructuredGridWriter in the vmtkimagewriter as the vti format is always axis aligned. Writing to vts format provides a means to have a format that can be viewed in paraview in the correct coordinate system

rlizzo:
This is resolved in #255

PR #255, opened 2018-03-29
add vtkXMLStructuredGridWriter to vmtkimagewriter

kayarre:
This addresses a work around in my workflow that bypasses the issue of #254 .
basically the vtk .vti format doesn't support generic orientation, only providing axis aligned output, however the structured grid format does.

Here is an example of how one would use this writer:
vmtkimagereader -ifile VWI_post.nrrd --pipe vmtkimagewriter -transform 1 -matrix @vmtkimagereader.xyztorasmatrix -ofile VWI_post_test.vts

vmtkimagereader -ifile VWI_post.nrrd --pipe vmtksurfacetransformtoras -ifile segmentedsurface.vtp -matrix @vmtkimagereader.xyztorasmatrix -ofile segmentedsurface_ras.vtp

VWI_post_test.vts and segmentedsurface_ras.vtp are both mapped to the same space when viewed in paraview. This is important in that segmentedsurface_ras.vtp is later used to create an unstructured volume mesh to compute flow. This provides a means to visualize/post process the image volume with the flow information in paraview.

### DIFF
--- a/vmtkScripts/vmtkimagewriter.py
+++ b/vmtkScripts/vmtkimagewriter.py
@@ -44,7 +44,7 @@ class vmtkImageWriter(pypes.pypeScript):
         self.SetScriptDoc('write an image to disk')
         self.SetInputMembers([
             ['Image','i','vtkImageData',1,'','the input image','vmtkimagereader'],
-            ['Format','f','str',1,'["vtkxml","vtk","meta","tiff","png","pointdata"]','file format'],
+            ['Format','f','str',1,'["vtkxml","vtk","meta","tiff","png","pointdata", "vtsxml"]','file format'],
             ['GuessFormat','guessformat','bool',1,'','guess file format from extension'],
             ['UseITKIO','useitk','bool',1,'','use ITKIO mechanism'],
             ['ApplyTransform','transform','bool',1,'','apply transform on writing - ITKIO only'],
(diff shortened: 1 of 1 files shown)


lassoan:
The proposed changes in this pull request are a bit unclear to me, but I have the feeling that such things should not be necessary. 

ParaView ignores axis directions and vti files cannot even store it. If you are keen on using Paraview then I think the best is to remove image orientation right after DICOM import and only work with axis-aligned images from that point. You lose original image coordinates, but if you really need a mapping then you can save the alignment transform.

These extra steps are clearly not convenient and that's why we essentially stopped using Paraview for medical image computing applications and implemented necessary features in 3D Slicer instead. I would only recommend Paraview if you only work with mesh data. If you want to visualize or analyze images as well, then 3D Slicer is much more suitable. If there is anything that you do in Paraview and not sure how to do in Slicer then let me know (or post it to discourse.slicer.org).

kayarre:
@lassoan thank you for the helpful feedback. I cleaned up my description to be more specific to the work around. I agree with you about using 3D Slicer, and will plan on doing more there as I am able.

lassoan:
Thank you, I understand it now. It could be useful to describe the rationale in the code as well, in method documentation and maybe in comments how and why VTK structured grid can be used for volume storage and visualization. Probably it should be described explicitly that it is for allowing rotated volume visualization in Paraview - there would be probably not many other uses, since only a handful of software applications can visualize this kind of data, and even in those it is only for visualization, since no image processing algorithms can be executed directly on this kind of data.

rlizzo:
Thanks for the contribution Kurt! (and the great feedback Andras!)

I absolutely agree that we should aim to support all the different workflows, and transformations required for them, this is a welcome addition to the library. However, Andras made some great points, and I can see how the day-to-day utility of this feature is fairly opaque. Before we go ahead and merge this PR along with https://github.com/vmtk/vmtk/pull/253, we should include some more specific documentation (both in code comments, and in the docs on the [vmtk website](www.vmtk.org))

I think this whole topic would be ideal to include in a tutorial on the [tutorials](http://www.vmtk.org/tutorials/) webpage. The whole discussion of "what to do once you've done some work with vmtk" is hardly covered (though there have been some posts on the mailing list). Would you be interested in collaborating on this with me?

There are two other concerns I have at the moment:
1) We also need a reader for the `vts` file format. This should be easy enough to implement in vmtkimagereader. 
2) Need to add some tests. 

Let me know your thoughts; I can assist on any of the items that you would like me to.

rlizzo:
Also, the PR title should be "vtkXMLStructuredGridWriter" rather than "vtkXMLUnstructuredGridWriter". Caused a bit of confusion on our end :)

PR #256, opened 2018-03-30
add flow extensions to all outlets except those indicated

kayarre:
add the option to exclude boundaries from adding flow extensions on and add them everywhere else.

Use Case: many outlets and one inlet, apply the same kind of flow extension to all the outlets while exclude adding one to the inlet.

#### vmtk to create flow extensions outlets
vmtk vmtkflowextensions -ifile case1_mra_vmtk_decimate_ras2_dsatrim2_ext.vtp  -centerlinesfile case1_mra_vmtk_decimate_ras2_dsatrim2_crtlines.vtp  -adaptivelength 1 -extensionratio 2.0  -transitionratio 0.75 -normalestimationratio 1 -boundarypoints 100 -interactive 1 -exclude 1 --pipe vmtksurfacewriter -ofile case1_mra_vmtk_decimate_ras2_dsatrim2_ext2.vtp

### DIFF
--- a/vmtkScripts/vmtkflowextensions.py
+++ b/vmtkScripts/vmtkflowextensions.py
@@ -43,6 +43,7 @@ class vmtkFlowExtensions(pypes.pypeScript):
         self.ExtensionMode = "centerlinedirection"
         self.InterpolationMode = "thinplatespline"
         self.Interactive = 1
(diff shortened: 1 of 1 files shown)


rlizzo:
This is great! Nice job Kurt! Merging now

PR #257, opened 2018-04-03
Image Volume Viewer Capabilities

rlizzo:
This is a nearly complete PR (needs testing and potential user interaction controls) which adds volume visualization capabilities as a native features of VMTK.

The script accepts standard vtkImageData, places it in a vtkVolume object with appropriate mappers depending on the users system hardware capabilities: GPU prefered, though OSPRay and RayCast algorithms are supported as hardware fallbacks (or at the request of the user). 

A nieve default is provided for the three required transfer functions:
- scalar color
- scalar opacity
- scalar gradient opacity

However, the user can supply any desired list of points which will construct a custom visualization. 

In addition, the "EnableImagePlaneWidgets" option automatically adds x, y, z, imagePlaneWidgets as overlays within the render window. These are aquired via piping a single the vmtkRenderer instance through both vmtkImageViewer and the new vmtkImageVolumeViewer classes. 

Examples
---
Example of nieve volume rendering
```vmtkimagereader -ifile foo --pipe vmtkimagevolumeviewer```

image

---
Example of Overlayed image plane widgets
```vmtkimagereader -ifile foo --pipe vmtkimagevolumeviewer -enableplanewidgets ```

image


rlizzo:
@lantiga, please review at your convenience. 

Since the initial post, there have been major changes to the structure of the module. 

1) I removed the built in utility to create a render window with the overlayed image planes, as this functionality is already provided by piping vmtkRenderer appropriately. 

2) A wide variety of sensible defaults for the color, opacity, and lighting values / transfer functions are now provided. These default properties are taken from the 3D slicer volume render presets, with slight simplifications made to the XML data fields. As there was no obvious place to include the xml document, I created a new "resources" folder within vmtkScripts and have made appropriate modifications to the build system in order to place the document in the correct location at install time.

3) In order to allow user inspection / storage of parameters used in the visualization, the volume render properties are now provided as pype output members. 

4) a lot of comments and refactoring.. 

Let me know your thoughts.

rlizzo:
Made suggested changes and all tests pass. Merging now.

PR #258, opened 2018-04-04
Add Interactive Surface Selection

rlizzo:
This module allows a user to interactively select which non-connected components of a surface they wish to keep. 

At the current time, this actually works as intended. It just needs some refactoring and tests

Demonstration video is below:

https://www.dropbox.com/s/b37zkbwm6vgokkp/surface-selector-demo.mov?dl=0

### DIFF
--- a/vmtkScripts/CMakeLists.txt
+++ b/vmtkScripts/CMakeLists.txt
@@ -113,6 +113,7 @@ SET(SCRIPTS_SRCS
   vmtksurfaceclipper.py
   vmtksurfacecliploop.py
   vmtksurfaceconnectivity.py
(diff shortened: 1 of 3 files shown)


rlizzo:
Thanks for the review @lantiga! 

I've refactored the code with both performance and cleanliness in mind, and see significant performance increases. With the new method,vtkPolyDataConnectivityFilter is ran only once to extract and label all regions. From there it is called again to extract only the regionId which was selected by the user. This is added to it's own actor in the scene (rather than adding every component as individual actors). By storing each selected actor in a list, this also allows us to perform unlimited selections and 'undos'. 

This code is working as intended on python 2 and 3. am merging now!

PR #260, opened 2018-04-06
add 'all' feature to surface connectivity with surface tags

kayarre:
Sometimes the largest region is not what you want or maybe you need multiple surfaces.

The nice thing is that this will tag each connected surface which can then be filtered in paraview.


rlizzo:
Is this possibly related to https://github.com/vmtk/vmtk/pull/258?

kayarre:
#258 is a great feature addition! 
to some extent this a work around until #258, so it's certainly related and I am less sure what to do with in view of that. I still like that it adds functionality that already exists in vtk.

rlizzo:
I agree that this adds functionality, and that #258 (while related) achieves a slightly different goal. Seems like a useful addition! Merging now, thanks Kurt!!

PR #261, opened 2018-04-06
add vmtksurfacetomesh script

kayarre:
vmtk has vmtkmeshtosurface so why not vmtksurfacetomesh.

if one one wants to create a boundary layer mesh from a single surface the input must be an unstructured grid. This simple script provides the means to convert the surface to the unstructured mesh.

### DIFF
--- a/vmtkScripts/contrib/CMakeLists.txt
+++ b/vmtkScripts/contrib/CMakeLists.txt
@@ -21,6 +21,7 @@ SET(CONTRIB_SCRIPTS_SRCS
     vmtknumpywriter.py
     vmtksurfaceresolution.py
     vmtksurfacetonumpy.py
+    vmtksurfacetomesh.py
     vmtksurfacewriter2.py
     vmtksurfaceextractannularwalls.py
     vmtksurfaceextractinnercylinder.py
--- a/vmtkScripts/contrib/vmtksurfacetomesh.py
+++ b/vmtkScripts/contrib/vmtksurfacetomesh.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+## Program:   VMTK
+## Module:    $RCSfile: vmtksurfacetomesh.py,v $
+## Language:  Python
+## Date:      $Date: 2018/04/06 12:09:46 $
+## Version:   $Revision: 1.8 $
+
+##   Copyright (c) Luca Antiga. All rights reserved.
+##   See LICENCE file for details.
+
+##      This software is distributed WITHOUT ANY WARRANTY; without even 
+##      the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR 
+##      PURPOSE.  See the above copyright notices for more information.
+##
+## Contributed by Kurt Sansom
+
+from __future__ import absolute_import #NEEDS TO STAY AS TOP LEVEL MODULE FOR Py2-3 COMPATIBILITY
+import vtk
+import sys
+
+from vmtk import pypes
+
+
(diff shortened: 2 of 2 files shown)


rlizzo:
Hey Kurt,

The contribution is really appreciated, but i think that this is actually better left for advanced users to implement on an as needed basis. Creating an empty mesh container (consisting only of a surface) is not something that we really want to encourage. 

The unstructured grid container is rather space inneficient (storing point locations, non-compacted cell connectivity info, and cell shapes before even talking about data arrays). It's also noticably more difficult to work with than the analagous polydata API. If you don't need to use it, it's better off not to. 

Is there a use case that you are thinking of where this would be required? I think your recent contributions have pointed towards a certain workflow that is lacking functionality.

kayarre:
Thank you Rick for the feedback.
I think the main reason is that in order to use vmtkboundarylayer the input
says it requires a surface but that surface is required to be an
unstructured grid. My first attempt was to make that script be able handle
a polydata and unstructured grid. maybe that is the best place for it, but
it seemed much simpler and straightforward to create a separate script.

This was more of an experiment rather than a specific workflow that led me
to find The gap in functionality.

Regards,

-Kurt

kayarre:
Thank you Rick for the feedback.
I think the main reason is that in order to use vmtkboundarylayer the input
says it requires a surface but that surface is required to be an
unstructured grid. My first attempt was to make that script be able handle
a polydata and unstructured grid. maybe that is the best place for it, but
it seemed much simpler and straightforward to create a separate script.

This was more of an experiment rather than a specific workflow that led me
to find The gap in functionality.

here is the code from the vmtkboundarylayer script
```
self.SetScriptDoc('create a prismatic boundary layer from a surface mesh and a set of vectors defined on the nodes')
self.SetInputMembers([['Mesh','i','vtkUnstructuredGrid',1,'','the input mesh','vmtkmeshreader'],
```
it expects vtkUnstructuredGrid for the surface mesh.

rlizzo:
Out of curiosity, what happens when you feed in a unstructured grid created in that fashion to the bounary layer generator? Does it create one?

The mesh processing module is one Ive had very little input towards, I don't know if it expects volumetric elements to work or not.

kayarre:
I think its pretty neat in that it's using something like vtkWarpVectors to propagate the points along the normals and then building new cells from the surface/grid that is already there. however, vmtkboundarylayer doesn't seem to work very well when the layer is too thick and curvature is high.
I think essentially you get intersecting normals and it doesn't handle that very well from the minimal exploring I have done, as it tries to do mesh untangling.

rlizzo:
Yeah. I've ran into a similar problem before where I had to dynamically shrink/expand the radius of a tubular surface with a factor based on it's original radius (which varied drastically). When you try and warp the points along a normal they start to intersect, but the cell connections stay the same which completely messes up the geometry. 

I never actually published the study, but I still have the code all ready to go. Is this something that you might need for a project you're working on? 

Before:
screenshot 2018-04-12 18 59 25

After:
screenshot 2018-04-12 18 59 58

kayarre:
@rlizzo I just realized that this script already exists. I am baffled how I missed it and why I reinvented the wheel here. I will close this pull request pending additional comments.

Also I would be really interested to see your solution to the intersection problem.

rlizzo:
Wow just got that message on the mailing list as well! Learn something new about this tool everyday! 

Whats your email? I don't want to post the code publicly (yet) for a number of reasons

kayarre:
@rlizzo sent you an email to the one listed on the git log, mine is there as well.
script already exists closing

PR #262, opened 2018-04-09
Updated Readme

rlizzo:
Updated readme so that the features of the library are included in the
main github repo readme. These are taken from the vmtk.org webpage with
the html formating simplified to the equivalend markdown format.


rlizzo:
merging as just a doc update

PR #265, opened 2018-04-12
[WIP] Automatic Centerline Extraction

rlizzo:
This is an early stage PR which is designed to implement an automatic flux-driven medial curve centerline extraction method originally published at: http://www.insight-journal.org/browse/publication/165

The unmodified source does compile and execute as expected with ITK version 4.13.0 and C++11 (only tested on mac). While the core algorithms seem solid, there are a number of modifications which are needed in order to integrate it into VMTK:
1) Implement VTK setters/getters for the data pipeline.
2) Convert a surface to an image containing the signed distance map (the expected input for the algorithm). 
3) Convert the output binary image (which specified centerline points) to a polydata centerline representation. 

This code does not compile at the current time, but I am posting this to get feedback on the method.


rlizzo:
Since I came up with a much better method in #270, I'll close this PR. @lantiga, do you think any part of it is work keeping around for use? I think [vmtsurfacetobinaryimage](https://github.com/vmtk/vmtk/pull/265/files#diff-42242af3476247108603cbff9d622cd1) might be?

rlizzo:
With the differences in the way that ITK compiles on linux, mac, and windows, the results of this script are slightly different on the different platforms. I have updated the helptext of `vmtkcenterlineimages` to reflect that. I have also included test data for the output on the three major supported OSs (and the tests pass). 

This PR adds two new scripts.
1) **vmtksurfacetobinaryimage**
```
vmtksurfacetobinaryimage : Convert a surface to a binary image mask
  where voxels are labeled if they lie inside the surface or not.
  Input arguments:
   -id Id (str,1); default=0: script id
   -handle Self (self,1): handle to self
   -disabled Disabled (bool,1); default=0: disable execution and
     piping
   -i Surface (vtkPolyData,1): the input surface
   -ifile SurfaceInputFileName (str,1): filename for the default
     Surface reader
   -invalue InsideValue (int,1) >= 0 and <= 255; default=1: the value
     to label voxels which contain points inside the surface
   -outvalue OutsideValue (int,1) >= 0 and <= 255; default=0: the
     value to label voxels which contain points inside the surface
   -imagespacing PolyDataToImageDataSpacing (float,3) >= 0.0;
     default=[0.3, 0.3, 0.3]: the sample spacing for the polydata to
     image data conversion
   -ofile ImageOutputFileName (str,1): filename for the default Image
     writer
  Output arguments:
   -id Id (str,1); default=0: script id
   -handle Self (self,1): handle to self
   -o Image (vtkImageData,1): the binary surface image
```

2) **vmtkcenterlineimage**
```
vmtkcenterlineimage : Automatically extract a binary skeleton
 centerline image from a surface using a flux-driven medial curve
 extraction algorithm. Note: results will be slightly different on
 different operating systems.
  Input arguments:
   -id Id (str,1); default=0: script id
   -handle Self (self,1): handle to self
   -disabled Disabled (bool,1); default=0: disable execution and
     piping
   -i Surface (vtkPolyData,1): the input surface
   -ifile SurfaceInputFileName (str,1): filename for the default
     Surface reader
   -sigma Sigma (float,1) >= 0.0; default=0.5: the kernal width of the
     gaussian used for image smoothing
   -threshold Threshold (float,1) >= 0.0 and <= 1.0; default=0.0: the
     threshold to filter the output skeleton image at
   -spacing PolyDataToImageDataSpacing (float,3) >= 0.0; default=[0.3,
     0.3, 0.3]: the sample spacing for the polydata to image data
     conversion
   -ofile ImageOutputFileName (str,1): filename for the default Image
     writer
  Output arguments:
   -id Id (str,1); default=0: script id
   -handle Self (self,1): handle to self
   -o Image (vtkImageData,1): the output centerline image
```

I am merging this now since it achieves the purpose. Screenshots of the final output are below:

**Example 1: Simple Case**

image

image

**Example 2: Complex Case**

image

image

Issue #248, opened 2018-03-01
Linking External ITK and VTK doesn't force rebuild after change


kayarre:
when linking external ITK and  VTK making changes to vmtk doesn't force a rebuild

have to ` make clean; make`

rlizzo:
This is a behavior i've been annoyed by for a long time. I don't really understand it since the vmtkScripts/cmakeLists.txt file install logic should identify the python files which have changed and automatically reinstall them when running `make`. The relevant code is here:

https://github.com/vmtk/vmtk/blob/0be71a3546c3f0d063ee6ae5f033798f393e10d5/vmtkScripts/CMakeLists.txt#L169-L171

@lantiga, any ideas on this one?

rlizzo:
It gets wierd with the C++ parts. Ive been playing around with it a bit but
have noticed some odd behavior that (only sometimes) if you only edit the
header files it won't find anything to update. It seems to find changes
every time with the .cxx files.

However, updates to the python scripts never update on edits without
running "clean" first.

The forums haven't been particularly helpful here either. This sort of
thing has been reported to occur when you use a "FILE ( GLOB foo)"
statement, since apparently that is only executed at generation time. The
general advice is to add a cmake_custom_command to pick up changes, but
it's not really apparent how to this would solve the problem or be relevant
to our build system. I'm working on a c++ feature addition now, so maybe
ill stumble across a solution to it. If not I'll probably post on the CMake
mailing list to see if I can get any other info.

rlizzo:
Actually, I was wrong, the C++ side only rebuilds if there is a compilation error. otherwise you have to `make clean` and `make` in order for it to pick up changes.

lassoan:
Do you expect running `make` to install Python files? For installation, you need to run `make install`. If you need to copy files during build step then you need to add a custom target for that.

lassoan:
I had a look at the code. Probably the issue is that you copy many files using CONFIGURE_FILE.

rlizzo:
Wow. this was actually a really tricky behavior to understand since it's actually not an issue with the CMakeLists.txt files, but with how we use VMTK as a superbuild project. However, it's fixed by specifying a single parameter :) 

PR: https://github.com/vmtk/vmtk/pull/267.

From the last paragraph of page 17 in [this article](https://blog.kitware.com/wp-content/uploads/2016/01/kitware_quarterly1009.pdf). 

[...]

PR #267, opened 2018-04-13
Fixes rebuilding VMTK after change to source files.

rlizzo:
When the CMake option "USE_SUPERBUILD" is enabled, CMake treats VMTK as
an External Project which pulls its source from the local disk. By
building in this manner, the CMake configures both the compilation and
instalation step to occur at the same time with a single call to "make".

While this works when a project is built from source (the first time),
the default behavior of an external project is to not feed in the source
files as a dependency of the generated install. As a complete list of
header files is not given as input to the build step, there is nothing
to indicate that the build step should re-run. Any changes which are
made to the source directory are therefore not rebuilt in the
installation directory.

By setting the CMake External Project option BUILD_ALWAYS 1, a build
timestamp file is not generated by cmake, and this external project
always builds, therefore picking up any changes in the source.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -181,6 +181,7 @@ ExternalProject_Add( ${proj}
   DOWNLOAD_COMMAND ""
   SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}"
   BINARY_DIR VMTK-Build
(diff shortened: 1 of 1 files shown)


rlizzo:
Addresses https://github.com/vmtk/vmtk/issues/248

PR #268, opened 2018-04-24
[WIP] PypeS Refactor & Documentation

rlizzo:
Hey @lantiga, in preparation for trying to integrate pype execution into jupyter notebooks, I started to really dig into the PypeS package. As I read through it, it was very apparent that the code was in need of a bit of TLC (most of it hasn't changed since you switched over from svn to git 11 years ago). This PR is the work that I've done just to try and understand/simplify what is happening when a pype is executed.

I originally just wanted to add docstrings and tests, but the complexity of the PypeS execution model meant I spent a lot of time working with the code just to understand what was going on. This turned into a bit more of a refactoring project that I intended.  So far I've tried to (and hopefully succeeded at) keeping the class boundaries and core execution structure intact.  Mostly I've just worked to simplify expressions, remove redundant or unused code, and extract individual units of work from large method blocks into their own functions. 

It's a work in progress, and not really ready for a thorough review, but I wanted to put this up here so I could put it on your radar and ask a few questions along the way.


rlizzo:
I should note that while this seems to work well in manual testing (atleast on my mac), the automated tests of the PypeS package are lacking and will need to be improved before I'd feel comfortable merging.

PypeS/pype.py
```diff
@@ -144,7 +166,7 @@ def ParseArguments(self):
                 self.AutoPipe = 0
             if '--nolog' in pypeArguments:
                 self.LogOn = 0
-            if '--query' in pypeArguments:
+            if '--query' in pypeArguments: #TODO: What on earth does this do?
```
lantiga: If I remember well 😄 , it was a facility to get all scripts from a package. Mainly done for PypePad. It's broken now, it should go for good.

lantiga:
As we think about moving this stuff into Python, we should figure out what do to with the `@` sigils, ids and such. Let's remember to talk about this today.

rlizzo:
Thanks luca, I have been testing the pushed arguments (that took me a while to figure out how it was working in the code, I'll definitely document it), and everything seems to work as expected.

The one thing which is really bothering me is how method names are duplicated in `pypescript.py` and `pype.py`. It get's confusing when you have `ParseArguments()`, `Execute()`, `PrintError()`, and `PrintLog()` methods in both classes (especially since pype.Pype operates on a pypeScript object 😄 )

Looking forward to our discussion, I definitely have a better handle on how you're thinking this will work.

lantiga:
Yeah, so the reasoning is that Pype is the composition of one or more PypeScripts. So e.g. printing or parsing is something that the parent (Pype) has to do to parse the pype-wide args and break up the whole pype into calls to scripts, and then each PypeScript knows how to parse, etc its own script-specific option.
So the name duplication was meant to be intentional (which doesn't make it less confusing of course), to convey this two-layered structure.

rlizzo:
Well that certainly makes sense, but there's got to be a simpler way to handle that. I'm sketching out a few ideas now for a refactor, and will propose them as I work through them. I'm really not a fan, since it's not clear at all where a message is being printed from (unless you have intimate knowledge of how the pypes system works).

rlizzo:
Since this will be a fairly large project, I'm going to create a seperate develop branch in the vmtk repo to direct all pull requests towards. It'll be easier to parse in the future if all this work (plus the work to come) is not performed in one giant github PR/branch.

rlizzo:
going to merge this into the `develop-pipes` branch now.

PR #270, opened 2018-04-27
[WIP] Network based automatic centerline

rlizzo:
### Summary
This PR is meant to replace https://github.com/vmtk/vmtk/pull/265 because the process of reconstructing a line from a collection of points in 3D space is an unreliable method, and the results would not have been that great. 

This method is just a clever way to compose the existing functionality of vmtk in order to automatically generate centerlines of any type of tubular structure (both branching and cyclic). It has surprisingly effective results. The code is well documented, but a brief overview of the method is below. 

### Method
1) Cap any holes (if they exist) in the surface. 
2) choose a triangle cell from the surface at random, delete it. We now have a hole in a random location. 
3) use `vmtknetworkextraction` on the output of (2) to extract a network of approximate centerlines along with the network topology array. 
4) convert the network (3) from `vtkPolyData` to a `numpy array` (just for ease of editing).
5) remove the centerline which started at the random hole we created in (4). Since it was a random cell, it will most likely be a short segment which intersects a tube at 90 degrees. 
6) precompute the `delaunay triangulation` of the capped surface (1) (to expedite future processing). 
7) for every segment in the network topology array (5), determine the start and end points (xyz locations) of the segment
8) use vmtkcenterlines non interactively with the `pointlist` option enabled and the `sourcepoint` and `endpoint` set to the points determined in (7). Feed in the capped surface (1) and delaunay triangulation (6). store each resulting centerline in a list.
9) create an empty vtkPolyData object, and append every centerline segment within (8) to it. 

### Bugs and ToDo List
1) ~~the `EdgeArray` is not being property converted back from a numpy array to a point data tuple; which is probably a bug in `vmtknumpytocenterlines`.~~
2) ~~the standard out window is being heavily polluted by a ton of output from repeatedly calling `vmtkcenterlines`, `vmtkcenterlinestonumpy`, and `vmtknumpytocenterlines`~~
3) ~~It's not clear if this should reside in the main `vmtkcenterlines.py` script or in it's own thing~~
4) I haven't run any formal comparisons

### Performance
Serially executing the algorithm is rather slow. This is because `vmtkcenterlines` has to be called for every segment extracted from the network. For a simple case (case 1 below), there are 30 individual segments. For a more complex case, there are upwards of 100 individual segments. Precomputing the delaunay triangulation does speed up execution of `vmtkcenterlines` by about 70% in my tests, but even with that speedup each centerline iteration take 10-15 seconds. Even though it's completly automatic, complexity can make the processing time quickly increase to an inconvenient amount of time for a user to wait. 

As each segment's calculation is independent of another, I decided to use the [Joblib](https://pythonhosted.org/joblib/index.html) library to execute each centerline calculation in Parallel, utilizing all available CPU cores.

On unix-like OS's, joblib works by forking the parent process into independent processes and managing the delayed execution of some function across the input collection. However, joblib (and really all python parallelization libraries) require serializable functions and input arguments/return value, which is an issue since a `vtkDataObject` is not serializable. Instead of converting the inputs to numpy `ArrayDict`'s and having to reconstruct the `vtkPolyData` each time the loop executes (a slow process for surfaces), I decided to just pass in the memory address of the data objects as a string, and use the vtk python bindings to create a python name referring to the data residing at that memory address in each fork. While the forks can access the memory space of the parent process, the parent process cannot access those of the children, and as such we cannot use the same pass by reference methodology for the return values. We have to convert to a numpy `ArrayDict` in the children, and then reconvert back when control resumes in the parent. 

I have no ungodly idea how this will play out on a windows system, and am sortof expecting to run into issues since we are no running the loop from the __main__ block (see the warning here: https://pythonhosted.org/joblib/parallel.html). I'm also a bit worried about controlling garbage collection, since if the delaunay tessalation or surface objects are collected before (or in the middle of) executing the parallel loop, the program will segfault as it tries to access memory that's now gone out of bounds. It hasn't been an issue yet, but I'm considering explicitly disabling garbage collection in this segment. It's not a great practice, but it is probably better safe than sorry here. 

#### Other Impacts
We now need to specify joblib as a dependency of the project. This is a very common package, and is available both on the anaconda core and conda-forge channels, so no worries there. 

#### Case 1 - Simple Structure
image

image

#### Case 2 - Complex, Cyclic Structure
image

image

Looking forward to hearing your thoughts.


rlizzo:
@lantiga, after our discussion I went and did a basic profile of `vtkvmtkpolydatacenterlines.cxx` to see where the big holdup was. When I fed in the `DelaunayTriangulation`, the algorithm took about 15seconds to run. The time spent in each of the functions is listed below:

TotalTime -> ~15 sec
1) `vtkPolyDataNormals` -> ~ 0.5 sec
2) `vtkvmtkVoronoiDiagram3D` -> ~4.4 sec
3) `vtkvmtkNonManifoldFastMarching` -> ~9.5 sec
4) `vtkvmtkSteepestDescentLineTracer` -> ~0.16 sec

By making the `VoronoiDiagram` an optional input and acquiring the `PoleIds` from the `vmtkDelaunayVoronoi` script, I can completely remove `vtkvmtkVoronoiDiagram3D`. However, the `vtkvmtkNonManifoldFastMarching` is still a bit opaque to me. 

I'm doing some reading to understand the theory behind the method. Is there a specific reference/paper you based the implementation off of? If so, can you send it to me?

rlizzo:
Thanks for the feedback today @lantiga. I fixed the error with the `StopFastMarchingOnReachingTargetPoint` behavior by breaking the out the check for `StopSeedId` into it's own statement. 

I also did a brief comparison of the speedup we receive between a surface with ~30 independent segments and a surface with ~100 independent segments. (running on my 4 core, 8 threads mac)

30 Segments -> `StopOnReachingTarget == True` -> Serial Execution = 56 seconds
30 Segments -> `StopOnReachingTarget == False` -> Serial Execution = 210 seconds (3.5 min)
30 Segments -> `StopOnReachingTarget == False` -> Parallel Execution = 51.6 seconds
**30 Segments -> `StopOnReachingTarget == True` -> Parallel Execution = 14 seconds**

100 Segments -> `StopOnReachingTarget == True` -> Serial Execution = 150 seconds (2.5 min)
100 Segments ->`StopOnReachingTarget == False` -> Serial Execution =  1040 seconds (17.3 min)
100 Segments -> `StopOnReachingTarget == False` -> Parallel Execution = 130 seconds (2.2 min)
**100 Segments -> `StopOnReachingTarget == True` -> Parallel Execution = 34 seconds**
 
As you prediced, the speedup is pretty incredible once we reach higher branch counts! However, even with the speedup of stopping fast marching propagation once the target seed is reached, I think it's important that we continue to execute this in parallel. I've added Joblib as a runtime requirement to the anaconda package and documentation. If you're good with the additional dependency, I'll update the conda-recipes repo and website documentation for the `dev` version.

I also edited the behavior on widows the reflect the limitations of Joblib. If this runs on a windows platform, the centerlines extraction has to execute serially since we can't guarrentee that the pipe will be run from a `if __name__ == "__main__"` block. The behavior is to print a message to the user to let them know that this is why it takes so long.

**Before I go ahead and merge**, I'd just like you to give the code a once over. I'm also a bit torn on weather to keep this as it's own script, or if I should merge it with the main `vmtkcenterlines` script? If we do decide to keep it as its own, I'm not sure if the name fits as it. `vmtkcenterlinesnetwork` is hardly descriptive. Hoping you have some input on that?

rlizzo:
This is ready to be merged. Once the JOSS submission is finalized I will bring this into master. Don't want to change the runtime requirements in the middle of a review.

rlizzo:
Tests pass, ready to merge now!

PR #281, opened 2018-05-19
Add Multiple Seed Colliding Fronts Method to vmtkImageInitialization

rlizzo:
This is in response to a [question posted on the mailing list](https://groups.google.com/forum/#!topic/vmtk-users/9YghcHqjSc8) by @kayarre and commented on by @lantiga and myself. 

This PR attempts to extend the colliding fronts method to allow for multiple seeds to be selected at the same time. The PR currently works (with a few bugs in the seed colors). 

Instead of being limited to selecting two seeds at a time, the user is allowed to input any arbitrary amount of seeds (as long as the total number is even). Every two seeds selected are assumed to belong to a single colliding front "seed pair". The colliding fronts initial level set is calculated `numSeeds / 2` times, once for each seed pair. These are then summed together to produce the "final" (combined) initial level set which is set as the script output. 

Looking forward to your feedback, not sure if there is a easier way to accomplish this task.

### DIFF
--- a/vmtkScripts/vmtkimageinitialization.py
+++ b/vmtkScripts/vmtkimageinitialization.py
@@ -18,6 +18,7 @@ import sys
 import math
 import string
 import vtk
+from random import random
 
 from vmtk import vtkvmtk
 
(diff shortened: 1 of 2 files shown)


kayarre:
@rlizzo thank you for this. I haven't gotten around to testing it yet, but will in the next week.

lassoan (approved):
Even thought the new method is not tested and (almost) not documented, it may be useful and unlikely to cause any issues. It is better to get it merged to ensure that it does not become outdated on an unmerged branch.

PR #294, opened 2018-05-22
Fix TypeError raised by font size in vmtkrenderer

gabrielbdsantos:
Force [_GetScreenFontSize](https://github.com/vmtk/vmtk/blob/master/vmtkScripts/vmtkrenderer.py#L101-L133) to return an `int` value, avoiding a `TypeError` when trying to set the font size.

### DIFF
--- a/vmtkScripts/vmtkrenderer.py
+++ b/vmtkScripts/vmtkrenderer.py
@@ -121,7 +121,7 @@ class vmtkRenderer(pypes.pypeScript):
 
         baseFontScalePerPixelWidth = math.floor(baseScreenWidth / baseFontSize)
 
-        scaledFontSize = math.ceil(userScreenWidth / baseFontScalePerPixelWidth)
+        scaledFontSize = int(math.ceil(userScreenWidth / baseFontScalePerPixelWidth))
 
         # make sure that the font size won't be set too low or high for low/high screen resolutions. 
         if scaledFontSize < 8:


rlizzo:
Hi @gabrielbdsantos, thanks for the contribution! I'm curious though, how did this become an issue for you? The call to [_GetScreenFontSize](https://github.com/vmtk/vmtk/blob/163082fc63be39d31b94b70b35a29e5e5e687b56/vmtkScripts/vmtkrenderer.py#L101-L133) should currently ensure that the return value is an integer type:

https://github.com/vmtk/vmtk/blob/163082fc63be39d31b94b70b35a29e5e5e687b56/vmtkScripts/vmtkrenderer.py#L333

Under what circumstances did this throw an error? 

In any event, this change more sense than the current method. My only suggestion is to delete the (newly) redundant `int()` call around `self._GetScreenFontSize()`.

gabrielbdsantos:
I have to be honest. I was careless to submit this PR.

Well, I was running vmtk 1.4 installed via conda when I came across this issue. When using vmtkimageviewer or any other viewer, a TypeError was being raised every single time. I just took a quick look on the last commit to see if some solution had already been implemented, but I failed to notice that there was in fact a `int()` call somewhere else in the code. This had already solved the issue!

Closing it.

rlizzo:
Ok. thanks. If you can't use the package because of this that's a big issue! I'll try to put together a bugfix release in the next few days. Thanks for the report!

PR #273, opened 2018-05-02
[WIP] Pypes Refactor

rlizzo:
I am creating this PR in order to centralize conversation around the proposed changes to the Pypes Modeule, Jupyter Notebook Integration, and the general package structure layout. 

At the time of writing, the `develop-pypes` branch is composed of #272 #271 #268. 

More information to follow.

PR #297, opened 2018-05-24
WIP: add obj reader and writer.

kayarre:
Update the surface reader and writer to work with wavefront obj format.

vtk can now write ascii wavefront format "obj", https://gitlab.kitware.com/vtk/vtk/merge_requests/4045
adding obj format to writer for when vtk version gets updated, currently won't work unless vtk master is linked to vmtk.

vtk could already read wavefront format so this portion already works.

also making sure the extension of the file is compared in lower case. I have seen ".PLY", ".OBJ", and "STL" in the wild.


rlizzo:
Thanks Kurt! This will be a nice addition. Was the wavefront reader included in the VTK 8.1.1 release https://github.com/Kitware/VTK/releases? 

Since this is a very new featurein VTK, I think it's probably best to implement a check for the VTK version the user is running and if it's less than whatever release this falls into just print a message saying their current build is out of date. Better to do that then have VTK crash the interpreter with what I can only imagine would be a cryptic error message.

Also, Just as a matter of process were almost ready to merge a major update to the library structure in https://github.com/vmtk/vmtk/pull/273. The PypeS system needed a lot of TLC, and it turned into quite a probject. Nothing Central to the API is changing, and we've even added an awesome new vmtk.run() method, but it's been quite a task to modernize the package structure. Since we may have to sit on this for a few days anyways, would you mind rebasing off of `develop-pypes` and submitting it as a PR to that branch? This merge isn't going to be any easy one with all all the updates to master that I need to sync up. Getting this PR on the same history with the refactor should make it one less thing to worry about breaking. 

Thanks again!

rlizzo:
Also, if you want to give https://github.com/vmtk/vmtk/pull/281 a shot I'd be curious to see what you think. Was this what you were thinking of in the post to the mailing list the other day?

PR #304, opened 2018-05-30
 add wavefront reader/writer .obj extension

kayarre:
Update the surface reader and writer to work with wavefront obj format.

vtk can now write ascii wavefront format "obj", https://gitlab.kitware.com/vtk/vtk/merge_requests/4045
adding obj format to writer for when vtk version gets updated, currently won't work unless vtk master is linked to vmtk. (not in latest vtk release 8.1.1) will give an error message if vtk version is less than 9 for now.

vtk could already read wavefront format so that portion already works.

also making sure the extension of the file is compared in lower case. I have seen ".PLY", ".OBJ", and "STL" in the wild.

I believe this addresses the comments in #297 although I flubbed re-basing and did it a different way.


kayarre:
This can be merged after master is updated to build with VTK 9

kayarre:
pointing to wrong branch

PR #316, opened 2018-10-17
fix: skip cast if input image is already of type float

kayarre:
skip performing the cast operation in the event that the image type is of type float already.


kayarre:
@lassoan any thoughts on this?

lassoan (approved):
Thank you, it looks good.

PR #317, opened 2018-11-06
Fix build error with new VTK 8.2

lassoan:
Fix build errors when building with recent VTK versions.

1. VTK version was reversed from 9.x to 8.2
2. FileName member variable was renamed

All changes are backward compatible.

### DIFF
--- a/CMake/vtkMacroKitPythonWrap.cmake
+++ b/CMake/vtkMacroKitPythonWrap.cmake
@@ -80,7 +80,7 @@ endmacro()
 #!
 #!   Slicer_VTK_WRAP_HIERARCHY_DIR:
 #!
-#!     Directory where to ouptut the hierarchy files
+#!     Directory where to output the hierarchy files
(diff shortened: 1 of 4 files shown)


pieper:
👍

lassoan:
Travis-ci error is due to "UnboundLocalError: local variable 'removeCellLength' referenced before assignment" in vmtkcenterlinesnetwork.py. It happens only with Python 3.6 and clang. Does not seem to be related to the commits.

rlizzo:
Hey guys, I restarted the build. hopefully it was a one off thing and it'll pass this time. Will merge once i know what this issue is. sorry about the delay.

kayarre:
With the advent of the vtkmodules approach I am finding that VTK installs vtk in /lib/lib/ and therefore the vmtk_env.sh not longer works. does anyone have some insight on that? I think the work around is to use the paths in ./VTK-Build/unix_path.sh but this theory is untested.

PR #323, opened 2019-02-28
Update patch version of VTK and ITK

phcerdan:
Same major.minor version, but updated to latest patch version
fixing bugs.

For ITK, it nows compiles with gcc8.
For VTK, fix build error related to const conversion.


lantiga:
Thanks!

PR #330, opened 2019-06-19
Fix VTK license

nocnokneo:


lassoan:
Looks good to me. VTK license is indeed BSD - see https://vtk.org/about/#license

PR #340, opened 2019-10-08
fix export target error with newer cmake and deprecated throw warning

kayarre:
Export target break the vmtk build on ubuntu 18.04 with cmake 3.10.

additionally dynamic throw  has been deprecated.


lantiga:
Thank you @kayarre

PR #344, opened 2020-02-07
Fix build error with VTK 8.2

jhgoebbert:
Protected attribute "FileName" does not exist in class vtkDataReader in VTK 8.x any more:
https://vtk.org/doc/nightly/html/classvtkDataReader.html

Access to FileName, FilePattern, FilePrefix should be done by get functions in class vtkImageWrite even though here the protected attributes still exist in VTK 8.x in contrast to vtkDataReader:
https://vtk.org/doc/nightly/html/classvtkImageWriter.html


lantiga:
@jhgoebbert I'm coming back to this after a long time.
Although I don't have a lot of bandwidth, I hope to get things back into shape with your help.

I see that CI is not succeeding, it would be instrumental to have green run again. For the time being I'm going to merge this one.

PR #349, opened 2020-04-08
Clean up code for wall shear rate based on strain rate tensor

uschille:
Regression test pending.

### DIFF
--- a/vtkVmtk/Misc/vtkvmtkMeshWallShearRate.cxx
+++ b/vtkVmtk/Misc/vtkvmtkMeshWallShearRate.cxx
@@ -152,7 +152,7 @@ int vtkvmtkMeshWallShearRate::RequestData(
   double normal[3];
   double wallShearRate[3];
  
-  int i, j;
+  int i, j, k;
 
   if (!this->UseFullStrainRateTensor) {
     for (i=0; i<numberOfPoints; i++)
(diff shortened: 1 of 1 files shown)


uschille:
@lantiga I recommend to double check the sign convention. We would typically use a different convention, cf. https://doi.org/10.1007/s00791-017-0277-7. Perhaps Mehdi Najafi can comment.

lantiga:
@uschille thank you, I also noticed the sign issue. I think the important aspect is to make it consistent between the two definitions (normal only or full tensor).
Can you point out the main differences wrt #348 or even better target that PR directly? This way we can account for changes to the script and the class.

uschille:
> I think the important aspect is to make it consistent between the two definitions (normal only or full tensor).

@lantiga I agree. I have compared results for a simple test case and as expected the difference between the "normal only" and "full tensor" definition is small and the sign is consistent. @myousefi2016 is performing additional tests on a more realistic geometry.

The main differences wrt to #348 are related to code quality in `vtkvmtkMeshWallShearRate.cxx` 
- self-documenting variable names
- one statement per line
- stick to 1D array for tensors for consistency with declaration of `velocityGradient[]`
- use loops to compute tensor components (leave loop unrolling to compiler): improves readability and makes symmetry of strain tensor explicit; fewer lines of code to maintain
- make sign convention explicit in prefactor
You might still credit Mehdi Najafi for the WSR calculation.

lantiga:
Merging this one, thank you @uschille

PR #377, opened 2021-07-27
Fix deprecated numpy issubdtype calls

ramtingh:
np.issubdtype with int or float second argument was deprecated in numpy 1.14 (https://github.com/numpy/numpy/pull/9505). This changes the argument to np.integer or np.floating which should avoid issues with recent numpy versions.


lassoan (commented):
Thank you for your contribution.

At the time when this code was written VTK did not have numpy converters, but now it has:

https://github.com/Kitware/VTK/blob/master/Wrapping/Python/vtkmodules/util/numpy_support.py

Could you check if we could leverage VTK's converters here? (to reduce the amount of custom code that has to be maintained in VMTK)

lassoan (approved):
It would be better to use VTK's own numpy converters (https://github.com/Kitware/VTK/blob/master/Wrapping/Python/vtkmodules/util/numpy_support.py) but the change improves compatibility and does not have any drawbacks, so it is OK to merge. The code can be refactored later to take advantage of VTK's numpy converters.

PR #372, opened 2021-07-08
Add support for building against VTK 9.0 and remove support for VTK 7.1

jcfr:
### Fixes

* Fix `test_vmtkbranchextractor.py::test_cell_data_array_names/values`
* Fix use of fixtures in `vmtkbifurcationsections` and `vmtkbranchsections` tests
* Fix `vmtkcenterlinesnetwork.py`
* Fix `vtkvmtkSimplifyVoronoiDiagram` when built against VTK9
* Fix `test_vmtkimagefeatures` updating `itkUpwindGradientMagnitudeImageFilter`

### VTK 9 support

* Introduce option `VMTK_USE_VTK9` and remove option `VMTK_USE_VTK8`
* Java wrapping is not supported when building against VTK 9
* Changes adapted from Slicer/Slicer@c658b7b (ENH: Add support for building against VTK 8.2.0 or 9.0.0)
* `VTK_USE_FILE` is included only if VTK <= 8.90.

* To support building against both version of VTK and use the relevant names for targets and components, variables `VTK_TARGET_PREFIX` and `VTK_COMPONENT_PREFIX` have been introduced.

* Download vtkAddon CMake modules required to support wrapping against both VTK 8 and VTK 9

* Fix building of `vtkvmtkSimplifyVoronoiDiagram` against VTK 9 <s>by excluding the use of `vtkCellLinks::BuildLinks`. Following the introduction of Kitware/VTK@88efc80 (Reworking cell links classes for performance and consistent API), the API changed and the implementation need to be updated accordingly.</s>

* Adapt to `vtkPolyData` API changes introduce in VTK 9
 
* Update VTK 8 version from 8.1.2 to 8.2.0 to fix link error against Python 3

* Fix `-Wdeprecated-declarations` related to use of `vtkPolyData::GetPointCells`

* Fix VTK9 build conditionally using "const vtkIdType"
  
* Fix build error when `VMTK_USE_RENDERING` is ON

### ITK 5 support

* Add support for building against ITKv5 introducing `VMTK_USE_ITK5` option
* Upgrade VMTK to be compatible with ITK5 (:pray:  @lassoan)

### macOS support

* Fix packaging of Python libraries on macOS 
* fix: changes required for Xcode 10.2 / Mojave  (:pray:  @sbwilson)

### Build-system: Stylistic changes:

* Remove CMake-language block-end command arguments 
* Convert CMake-language commands to lower case

### Build-system: External project consolidation

* Consolidate passing of option to VTK external project
[...]

PR #385, opened 2021-11-07
Update buildsystem to support vtk9 fixes

kayarre:
changes based on @jcfr branch to get it compiling locally. I have not run tests yet.


kayarre:
I think is a good enough release candidate. I was able to get most of the tests working, disabled some, and updated the test data for others.  I continued to have some issues with how cmake finds python and therefore decided to update the cmake minimum to 3.15 and require python3. 

I also converted the test data repository to git-lfs and added a cmake flag so it won't `submodule init` everytime. 

`vmtkcenterlinesnetwork` has a weird bug that is not deterministic because it randomly picks a CellId to remove. The test was passing everytime until today and then randomly passed and failed.

kayarre:
I was able to fix the upwind image filter. I believe that gets all the tests running  that were expected to pass previously.

It appears there are a bunch of scripts used for the ITK5 migration [ITK5Preparation](https://github.com/InsightSoftwareConsortium/ITK/tree/master/Utilities/ITKv5Preparation) I think these might be worth running. Can someone who has done this give me some hints on using them. 

I think the `vtkvmtkSimplifyVoronoiDiagram` is still broken.
What are the other pieces that still need to be addressed?

kayarre:
@lassoan are these changes in the right direction or not? I don't know when or if i should expect feedback.

lassoan:
Thank you for all your work on this! Your changes look good to me, except I'm not sure if we should introduce submodules. Since github allows storing unlimited amount of test data as release assets and [CMake's ExternalData module](https://cmake.org/cmake/help/latest/module/ExternalData.html) can download these data sets (by setting the release name to the hash algorithm name, such as SHA256; and setting the filename to the hash value), there is no need to use submodules. Renaming file to its hash and uploading to a github release can manual but it is also quite easy to automate, as it is done in this repository: https://github.com/Slicer/SlicerTestingData.

If you are done with your changes then please submit a pull request to https://github.com/jcfr/vmtk/tree/update-buildsystem-to-support-vtk9. @jcfr will review and merge the changes, we'll test on Windows/Linux/macOS, then integrate the branch into VMTK master.

kayarre:
@lassoan thank you so much for the feedback. 

The test data was already a submodule, The difference is that I converted the submodule data to using `git lfs` instead of regular git. That seemed to make sense to me, but I know there are mixed experiences with git lfs usability.

I agree with how it is implemented is not ideal. Should I make an attempt at converting the submodule to using the cmake ExternalData?

I will go ahead and make a merge request to @jcfr.

lassoan:
@kayarre https://github.com/vmtk/vmtk/pull/372 is now merged. I think @jcfr included all your commits, except maybe 2 or so. Please check and either close this pull request and open a new one or rebase this one on the latest master. Thank you!

kayarre:
Awesome! I will take a look early next week.

kayarre:
Extracted relevant changes and put into new commits and merge requests. Closing

PR #391, opened 2021-11-29
fix: make test data submodule init optional

kayarre:
Make `submodule init` optional so test data can be selected or edited without overwrite during build process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,8 @@ option( VMTK_BUILD_TESTING
   "Build the VMTK testing libraries" OFF)
 
 if(VMTK_BUILD_TESTING)
+  option( VMTK_INIT_TEST_DATA
+    "Initialize the vmtk-test-data submodule" ON)
   add_subdirectory(tests)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,8 +102,6 @@ option( VMTK_BUILD_TESTING
   "Build the VMTK testing libraries" OFF)
 
 if(VMTK_BUILD_TESTING)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -102,6 +102,8 @@ option( VMTK_BUILD_TESTING
   "Build the VMTK testing libraries" OFF)
 
 if(VMTK_BUILD_TESTING)
+  option( VMTK_INIT_TEST_DATA
```
jcfr, line 105: To explicitly notify the developer of locally modified files, what do you think of explicitly checking for locally modified files using approach like this one [used in ExternalProject-gitupdate.cmake.in](https://github.com/Kitware/CMake/blob/065604b2b3cabef95c0f406c990274f439452bbd/Modules/ExternalProject-gitupdate.cmake.in#L152-L166).

If there are modifications, the call to `git submodule update --init` is skipped and a warning is reported.

If the directory is empty (and build testing is enabled), the call to `git submodule update --init` is always performed.

kayarre: I am fine with that approach, I had not considered it. It will take me some time to figure out and implement.

kayarre: Another thought is that if the test data is explicitly on a different  branch and not  limited to modified repo, we don't want to update it in that instance either. I am not sure the proposed method accounts for this, but will test.

kayarre: I updated things to use what you suggested. I tested that it does not overwrite the data in the unclean state.

kayarre:
@jcfr @lassoan is this good to merge?

lassoan (approved):
It looks good to me, thank you.

PR #392, opened 2021-11-29
fix: update reference data for otsu tests

kayarre:
update reference data for otsu tests.

https://github.com/vmtk/vmtk-test-data/pull/2


jcfr:
Could you confirm that the updated test data works when building against vtk8 ? 

If not, I suggest to:
* Rename current files including `_vtk8_` in the name (e.g `imagereference/test_vmtkimageotsuthresholds_vtk8_test_otsu_image_multiple_thresholds_0.mha`
* Add new files (no need to include `_vtk9_` as ultimately vtk8 support will be removed)
* Update test to conditional use these

kayarre:
I have not yet tested using with vtk8 or itk4. thank you for the feedback. I does make sense maybe to make different tests specific to the library version used if the test fail with the new data.

kayarre:
I think I have a plan to make vtk version specific tests, but have not tried that with pytest yet, will have to look into it more tomorrow.

kayarre:
@jcfr is this what you had in mind?
I finagled renaming the old files and add the new files for vtk9 and updated the test to conditionally run the tests.

kayarre:
@jcfr @lassoan is this good to merge?

lassoan:
Don't we need to merge https://github.com/vmtk/vmtk-test-data/pull/2 first?

kayarre:
Indeed you are correct.

kayarre:
This is ready once  [vmtk/vmtk-test-data#2](https://github.com/vmtk/vmtk-test-data/pull/2) is merged

kayarre:
@lassoan or @jcfr  any thoughts?

kayarre:
@lassoan I updated this MR to reference ITK5, tested on multiple variants, linted the python code.

I think the only piece that might be missing is a very thorough investigation to what commit made the change. I would be happy to do this, but with limited time and would appreciate guidance or help on how to do that efficiently.

kayarre:
@lassoan any thoughts on this?

lassoan:
> I think the only piece that might be missing is a very thorough investigation to what commit made the change. I would be happy to do this, but with limited time and would appreciate guidance or help on how to do that efficiently.

It was something about how the histogram bins are specified (switching between bin center and bounds). The difference should be negligible.

kayarre:
The difference is negligible but the test does not yet account for that. I would be up for a conference call sometime to discuss what the high priority things are and what makes sense design wise and scope out improving the tests.  Thank you @lassoan for your feedback and support.

per [vmtk/vmtk-test-data#2](https://github.com/vmtk/vmtk-test-data/pull/2) this is ready to merge

lassoan:
By negligible I meant that we don't need to further investigate the root cause and potential consequences of the change.

PR #393, opened 2021-11-30
fix: superbuild finds wrong python for VMTK build

kayarre:
VMTK fails to pickup correct python version for the superbuild.
this commit makes the variables use same ones for VTK super build.

I am open to suggestions if there is a better way, but it is a regular issue that I have where the python version is not picked up in external projects.


kayarre:
any thoughts on this? @jcfr @lassoan

lassoan:
We recently wanted to issue a similar fix in SimpleITK (see https://github.com/Slicer/Slicer/issues/5498). However, further investigation revealed that the problem is actually due to a bug in VTK (that it does not provide sufficient information for FindPython/FindPython3) and it was fixed here:

https://github.com/Kitware/VTK/commit/e54e756184412aea40c6915ccf31856f29b771d9

Could you please check if backporting this fix makes the problem go away for you?

kayarre:
I will give it a try.

kayarre:
@lassoan I am working on testing this out with `vtk 9.1.0`

if this fixes the issue we should update the superbuild to use `vtk 9.1.0`?

kayarre:
So far it looks like using `vtk 9.1.0` creates some additional errors though it does appear to fix this issue.

kayarre:
Closing this and make a new one with with fixes for `vtk 9.1.0` release

PR #395, opened 2021-12-06
STYLE: pep8 fixes

kayarre:
used script to catch fixes.

`$ for f in $(find . | ack "\.py$"); do \
  echo $f; \
  autopep8 --select W291,W292,W293,W391,E301,E302,E303,E304,E305,E306 --in-place $f; \
done`


lassoan (approved):
thank you

Issue #169, opened 2016-09-27
vmtksurfaceendclipper does no work consistently


kayarre:
when running the command:
vmtk vmtksurfaceendclipper -ifile simplification_fixed.ply -ofile simplification_fixed_clip.ply

I am also selecting points near the end of the model, I don't know if that effects things.

The command execute properly, but there is a unclipped triangle:
![screenshot from 2016-09-27 15-21-43](https://cloud.githubusercontent.com/assets/9031253/18894132/23231788-84c6-11e6-8d45-bd21bed90486.png)

I run the same command selecting points on a different branch and have different results.

vmtk vmtksurfaceendclipper -ifile simplification_fixed.ply -ofile simplification_fixed_clip.ply

> Done executing vmtksurfaceendclipper.
Writing PLY surface file.
ERROR: In /home/ksansom/build/vmtk/build/VTK/IO/PLY/vtkPLYWriter.cxx, line 122
vtkPLYWriter (0x49b3660): No data to write!
Output vmtksurfaceendclipper members:
    Id = 0
    Surface = vtkPolyData

kayarre:
It appears there are tow separate issues, The first is that the normal estimation routine gets an inverted normal sometimes and can't continue when the main body of the mesh  has been clipped. The second issue is the one I listed above where the triangles are fully clipped.

kayarre:
After further investigation I have been able to discern that the normal estimation filter often inverts the normal which leads to clipping the surface that you want to keep.

the second issue i believe is due to clipping an anisotropic mesh, meaning the seamfilter consistently creates a small triangle that does not lie on the implicit plane.

PR #399, opened 2021-12-20
BUG: Fix initial seam array value never assigned

a-mathieu:
Set `seamArray` value for the first element of the recursive loop instead of keeping the initialization value of 0.0 in `vtkvmtkTopologicalSeamFilter.cxx`

This should solve the extra triangle remaining issue with the `vmtksurfaceendclipper` script (#169). The initialization value of 0.0 for this specific point (top of the extra triangle) makes the clipping filter believe it belongs to the clipping plane resulting in an uncut triangle

Associated post on VMTK forum: https://discourse.slicer.org/t/extra-or-missing-triangle-with-vmtksurfaceendclipper-script/20530


kayarre:
@a-mathieu this is great! a while back I was struggling with this same issue but was not able to find a solution at the time. great work!

lassoan (approved):
Good catch, thank you!

PR #400, opened 2022-01-05
feat: Alternative surface boolean method

kayarre:
Add optional boolean method that provides faster, and potentially
  more robust boolean operation on poyldata

  includes additional test to verify functionality but currently
  excludes regression comparisons

### DIFF
--- a/tests/test_vmtkScripts/test_vmtksurfacebooleanoperation.py
+++ b/tests/test_vmtkScripts/test_vmtksurfacebooleanoperation.py
@@ -86,3 +86,19 @@ def test_operations_varied_tolerance(aorta_surface2, aorta_surface_reference,
     booler.Execute()
 
     assert compare_surfaces(booler.Surface, name, tolerance=1E-3) == True
+
+
+@pytest.mark.parametrize("operation,paramid", [
+    ('union', '0'),
+    ('intersection', '1'),
+    ('difference', '2'),
+])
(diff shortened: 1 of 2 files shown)


lassoan (approved):
VTK Boolean mesh operation implementations are all quite bad (I haven't found vtkLoopBooleanPolyDataFilter any better than vtkBooleanOperationPolyDataFilter), but since they are already in VTK it does not hurt to expose them. If anyone needs a much better, more robust VTK Boolean mesh operation implementation then there is https://github.com/zippy84/vtkbool .

kayarre:
Thanks @lassoan I have been waiting to attempt to bring in `vtkbool` hoping someone else would be able to get to it. Maybe it could happen in 2022.

lassoan:
vtkbool is available in 3D Slicer (provided by the Sandbox extension). It can be used from C++ and Python and there is also a module with a convenient GUI. So, you can at least test out everything before deciding to invest time into adding vtkbool as a dependency to the VMTK build system.

@jcfr is working on infrastructure that makes it easy to build, Python-wrap, and distribute VTK algorithm libraries. vtkbool will be one of the libraries that can be benefit from this. If you only need to use vtkbool in high-level Python scripts then pip-installing vtkbool may be an option (that is simpler to implement and maintain than adding vtkbool to VMTK SuperBuild).

PR #401, opened 2022-01-05
fix: Multiblock support to reader and writer

kayarre:
Add multiblock reader/writer support
  per: https://github.com/ppatalano/vmtk-petermods/commit/32b2bcccb4eda2f9c98cd42013382597a87073d5

Currently not tested


lassoan (approved):
Looks good, thank you. It would be nice if you could add a test in a follow-up commit.

PR #402, opened 2022-01-05
fix: Extract endpoint centerline sections

kayarre:
Adds script to extract endpoints from split and grouped centerlines.

  chore: move to contrib and pep8

  from: https://github.com/Clemson-MSE/vmtk/commit/1cff14b50fd75e06ef304a8d3398c93e3cb6571a

  fix: license and pep8

Currenlty not tested


lassoan (approved):
Thank you. It should be fine to bring in this new script. Even though it is untested and undocumented (other than the very terse self-description of the pypes script), it may be useful and unlikely to cause any issues. It is better to bring in useful features from forks so that they can be maintained in the official repository.

PR #403, opened 2022-01-05
WIP: Autoclip on previous mesh

kayarre:
Functionality for doing an auto clip function to repeat clips
  performed on previous mesh.

  from:
    https://github.com/ppatalano/vmtk-petermods

  commits:
    0f4e88768b6ee28264b5e260c82096a12bd85887
    806dfcf96dbd153f319bbeb81625792ec98edaee

  fix: move files to contrib
  chore: pep8
  fix: rename, license, and pep8

currently untested


lassoan (approved):
Thank you. It should be fine to bring in these new classes. Even though they are untested and undocumented (other than the very terse self-description of the pypes script), they may be useful and unlikely to cause any issues. It is better to bring in useful features from forks so that they can be maintained in the official repository.

PR #405, opened 2022-01-05
fix: vtk8 backwards compatibility for readers

kayarre:
This fixes the readers failing to compile due to changes in vtk 9.1

### DIFF
--- a/vtkVmtk/IO/vtkvmtkFDNEUTReader.cxx
+++ b/vtkVmtk/IO/vtkvmtkFDNEUTReader.cxx
@@ -52,6 +52,7 @@ vtkvmtkFDNEUTReader::~vtkvmtkFDNEUTReader()
     }
 }
 
+#if (VTK_MAJOR_VERSION >= 9 && VTK_MINOR_VERSION >= 1)
 int vtkvmtkFDNEUTReader::ReadMeshSimple(const std::string& fname,
                                        vtkDataObject* doOutput)
 {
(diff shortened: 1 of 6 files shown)


vtkVmtk/IO/vtkvmtkFDNEUTReader.cxx
```diff
@@ -52,6 +52,7 @@ vtkvmtkFDNEUTReader::~vtkvmtkFDNEUTReader()
     }
 }
 
+#if (VTK_MAJOR_VERSION >= 9 && VTK_MINOR_VERSION >= 1)
```
lassoan, line 55: This check will fail for VTK 10.0. Use VTK_VERSION_CHECK macro. Unfortunately, it is not available in VTK8, so it has to be added to VMTK if we really want to support VTK8.

lassoan (changes requested):
Thanks for the update. What is the rationale for supporting VTK8? Is there any actively developed project that needs it?

VTK8 and 9 are so vastly different that I don't think that keep spending time with backward compatibility is worth the effort. I've added a ticket to track this task: https://github.com/vmtk/vmtk/issues/406

vtkVmtk/IO/vtkvmtkFDNEUTReader.cxx
kayarre: Thank you for the feedback @lassoan. I ran into this issue testing something else.
I don't think it is necessary (at least for me), but wanted to put it out there to get feedback.

What is the way forward to deprecating the use of vtk8 with vmtk? 
What is the path forward to supporting vtk10?

lassoan: Actually, we could merge this as is (knowing that it will not work with VTK10), because after we merge this, we can create a 1.4 branch (that will be compatible with VTK8) and then drop VTK8 support from the master branch (and remove this VTK10-incompatible fix).

lassoan (approved):
Even though it is not compatible with VTK10, this is a temporary change (will be removed very soon), so it is acceptable.

PR #408, opened 2022-01-06
WIP: feature: contrib better loop clipping

kayarre:
Add contributed modified loop clipping to contrib


lassoan (commented):
There is a lot of code duplicated code with vmtksurfacecliploop.py. It would mean unnecessary increase in maintenance burden. Could you please merge it with the existing vmtksurfacecliploop.py script?

kayarre:
thank you @lassoan for the feedback I will investigate.

PR #412, opened 2022-01-18
Set random seed for non-deterministic test

ramtingh:
The behavior of `vmtkcenterlinesnetwork.py` is not deterministic since this script chooses a random node through python's `random` module, which means builds can randomly fail because of this one test. I think the easiest for now would be to use the same random seed for all builds, hopefully this at least makes sure the behavior is consistent.

(Technically different python versions can affect the random generator, but as far as I can tell it has been stable since python [3.2](https://stackoverflow.com/questions/11929701/why-is-seeding-the-random-generator-not-stable-between-versions-of-python))

### DIFF
--- a/tests/test_vmtkScripts/test_vmtkcenterlinesnetwork.py
+++ b/tests/test_vmtkScripts/test_vmtkcenterlinesnetwork.py
@@ -16,11 +16,12 @@
 
 import pytest
 import vmtk.vmtkcenterlinesnetwork as centerlinesnetwork
-
+import random
 
 def test_centerline_extraction_surface_with_no_hole(aorta_surface):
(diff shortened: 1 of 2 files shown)


lassoan (commented):
Thank you for working on this. If it matters what random values are generated then we cannot rely on the state of the global `random.Random` variable. Instead, we should fix the random generation in vmtkcenterlinesnetwork.py: vmtkCenterlinesNetwork should use its own `random.Random()` object and it should have an optional input parameter to set the seed value.


Setting the global random seed may interfere with anything and may be overwritten at anytime by anything, so we should not use it at all. Instead, could you please update vmtkcenterlinesnetwork.py to instantiate its own random generator object and add an optional seed parameter to ?

lassoan (approved):
Very good, thank you.

PR #414, opened 2022-01-22
Add CMake option to not expect test data to be a git repository

ramtingh:
CMake always tries to initialize the test data submodule, which can cause problems if I'm trying to build form a tarball or if git isn't available etc but still want the tests to run. 
This adds a `VMTK_TEST_SUBMODULE` option to disable this (defaults to replicating the existing behaviour)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,8 +85,11 @@ endif()
 option( VMTK_USE_SUPERBUILD
   "Build VMTK and the projects it depends on via SuperBuild.cmake." ON )
 
+option( VMTK_TEST_SUBMODULE
+  "Check and update VMTK Test data repository" ON )
+
 # Git setup
-if (VMTK_USE_SUPERBUILD OR VMTK_BUILD_TESTING)
+if (VMTK_USE_SUPERBUILD OR VMTK_TEST_SUBMODULE)
   if( NOT GIT_EXECUTABLE )
     find_package( Git REQUIRED )
   endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,8 +85,8 @@ endif()
 option( VMTK_USE_SUPERBUILD
   "Build VMTK and the projects it depends on via SuperBuild.cmake." ON )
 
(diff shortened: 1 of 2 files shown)


lassoan (commented):
This is very useful, thank you. My only concern is that the name of the option can be misleading (it suggests that it enables testing of some submodule).

Could you find a name that better describes its intent? For example for me these would be good enough:
- VMTK_USE_TEST_DATA_GIT_SUBMODULE
- VMTK_DOWNLOAD_TEST_DATA
- VMTK_TEST_DATA_SOURCE (this would be an enum with values such as `local` and `git-submodule` and could have further options in the future)

lassoan (commented):
Almost perfect, could you please specify the the accepted values? See example here:

https://github.com/Slicer/Slicer/blob/84d571bbc4358c8a64fca349d94dce12b0dee365/CMakeLists.txt#L526..L531

CMakeLists.txt
```diff
@@ -85,8 +85,11 @@ endif()
 option( VMTK_USE_SUPERBUILD
   "Build VMTK and the projects it depends on via SuperBuild.cmake." ON )
 
+set(VMTK_TEST_DATA_SOURCE "git-submodule" CACHE STRING "How to download test data (none, git-submodule)" FORCE)
+set_property(CACHE VMTK_TEST_DATA_SOURCE PROPERTY STRINGS none git-submodule)
```
lassoan, line 89: Sorry for being nitpicky, but it is probably better to clarify the details now than get back to it later.
What does "none" option mean? To me, source=none would suggest that no test data will be available (and so tests that require test data are skipped), but I don't think that the code does this (and then we could find a better name, such as `skip-data-with-tests`). If the actual behavior is to assume that the test data is already there then the option could be called something like "local" or "in-place".

ramtingh: I'll change it to `in-place`, to me local sounded like it might imply copying the data from some local folder. I essentially want an option that cmake can assume the data will be there and shouldn't download/check it.

lassoan:
Thank you, merged.

PR #415, opened 2022-01-23
bugfix: VMTK_TEST_DATA_SOURCE overriding command line

ramtingh:
@lassoan Sorry I've accidentally added `FORCE ` to this option and it needs to be able to be modified from the command line


lassoan (approved):
Looks good, thank you.

PR #424, opened 2022-04-14
Add an option to disable using joblib for network extraction

ramtingh:
[vmtkcenterlinesnetwork](https://github.com/vmtk/vmtk/compare/master...ramtingh:disable-joblib?expand=1#diff-687f0144e9a3d0c4c2cda087065c3db703fe15a9291a0136cac192a295e9029e) uses joblib for parallel computation, however it is somewhat clunky in that 1) there is no option to turn this off, 2) its not actually supported on windows, so for windows joblib is used for serial computation which just results in unecessary overhead compared to a for loop. 

This adds a `UseJoblib` option which defaults to true on osx/linux and false on windows. This shouldn't change current behaviour, but allows users to disable joblib if they don't want multithreading (due to weird behavior on conda-forge's build systems, I expect to have to disable joblib for testing on osx and this seemed like the cleanest way of facilitating it).


lassoan (approved):
Thank you, it looks good to me.

Issue #425, opened 2022-04-26
[Bug] vmtk window freezes


karabelaselias:
Dear all,
I operate Ubuntu 20.04 LTS with the nvidia-drivers 510 and cuda-10. First things first, I also use slicer which works fine and also the vmtk module operates well in it.

However, there are some things from vmtk I would like to use like the adaptive meshing and the boundary layers and the automatic surface capping. So I thought I need vmtk standalone

I compiled vmtk master according to the website and everything works out fine, however when I invoke the command

```
vmtksurfacereader -ifile foo.vtp --pipe vmtkcenterlines --pipe vmtkendpointextractor --pipe vmtkbranchclipper --pipe vmtksurfaceconnectivity -cleanoutput 1 --pipe vmtksurfacewriter -ofile foo_ct.vtp
```
as stated [here](http://www.vmtk.org/tutorials/SurfaceForMeshing.html) I run into the following issue:

First window pops up asking me to select seed points. I place one and press "q" to continue placing target points. However, the window refreshes and then freezes without any debug info in the console or apparent segfault (checked dmesg and syslog).

Has anyone else encountered a similar issue? Has this something to do with the nvidia drivers?

Best regards
Elias

isaiahchua:
Hi all, 

I run into the same issue using the same command both when running vmtk straight from the terminal and with python scripts. This only happens with the vmtk development version and not with the build installed from Conda. Any solution found for this issue? Thanks in advance.

OS: Ubuntu 20.04.2 LTS
Python version used when building vmtk: 3.9.12

Regards
Isaiah

ramtingh:
I see, The one hanging is ` vmtkcenterlines  ` with the default seed selector?

karabelaselias:
Exactly

ramtingh:
Hmm, I can't seem to reproduce that on windows at least. I'll try later on ubuntu

isaiahchua:
Hi @ramtingh , I've tried to install vmtk 1.5 using conda-forge as suggested on this page:
https://discourse.slicer.org/t/no-module-named-vtkmodules-vtkcommoncore/21854/7

These are the commands I ran to create the vmtk version 1.5 environments in both MacOS and Ubuntu:
`conda update conda anaconda-client`
`conda config --set restore_free_channel true`
`conda create -n vmtk2 -c conda-forge itk vtk vmtk=1.5`

[...]

PR #428, opened 2022-05-12
Add tests for undefined pypescript attributes

ramtingh:
pypeScripts declare command line options via SetInputMembers and expect these to be initialized separately, otherwise the command line version will fail while python bindings work fine (e.g. ` vmtkdistancetocenterlines ` in https://github.com/vmtk/vmtk/issues/425#issuecomment-1124520347) 
This test should catch this sort of issues


lassoan:
superseded by https://github.com/vmtk/vmtk/pull/494

PR #432, opened 2022-07-06
COMP: Remove deprecated/removed std::binary_function

sjh26:
std::binary_function was deprecated in C++11 and removed in C++17.
It only added typedefs. Since they were unused, it compiles fine
without them.

See: https://github.com/InsightSoftwareConsortium/ITK/commit/7b2dfd28d415d33c0138db275974735a9015fb30

binary_function is breaking the windows build on the Slicer extension factory now that we have switched to C++ 17, see:
https://slicer.cdash.org/viewBuildError.php?buildid=2733792

When this is merged, will update Slicer extension.


lassoan (approved):
Looks good to me, thank you.

PR #438, opened 2022-09-25
Update itkAnisotropicDiffusionVesselEnhancementImageFilter.h

chir-set:
Building SlicerVMTK fails without this patch.

At [cdash](https://slicer.cdash.org/buildOverview.php?project=SlicerPreview&date=2022-09-25) : 

> Errors:
> 31153-SlicerVMTK-git5458f53-g++-64bits-Qt5.15-Release: /.../SlicerVMTK-build/VMTK/vtkVmtk/Segmentation/itkAnisotropicDiffusionVesselEnhancementImageFilter.txx:409:6: error: no matching function for call to ‘itk::AnisotropicDiffusionVesselEnhancementImageFilter<itk::Image<float, 3>, itk::Image<float, 3> >::ResolveTimeStep(std::vector<double>&, std::vector<unsigned char>&)’
>

### DIFF
--- a/vtkVmtk/Segmentation/itkAnisotropicDiffusionVesselEnhancementImageFilter.h
+++ b/vtkVmtk/Segmentation/itkAnisotropicDiffusionVesselEnhancementImageFilter.h
@@ -228,9 +228,12 @@ private:
     TimeStepType TimeStep;
     std::vector< TimeStepType > TimeStepList;
 #if ITK_VERSION_MAJOR > 5 || (ITK_VERSION_MAJOR == 5 && ITK_VERSION_MINOR >= 3)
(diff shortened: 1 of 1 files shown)


lassoan:
Awesome, thank you for fixing this!

Could you tweak the implementation so that bool is only used for the latest  ITK versions (for previous versions uint8 was the correct type and it would be nice to keep compatibility with those versions as well).

chir-set:
> Could you tweak the implementation ...

Would this last update do ?

Please note that [this comment](https://github.com/chir-set/vmtk/blob/b2ceeb4b6cd74e4212008bf76d3ee5e69c505cae/vtkVmtk/Segmentation/itkAnisotropicDiffusionVesselEnhancementImageFilter.h#L233) won't be accurate anymore.

PR #453, opened 2023-04-12
Add missing attributes to vmtkdistancetocenterlines.py

JaroslavHron:
Add `UseRadiusThreshold` and `RadiusThreshold` atributes to `vmtkdistancetocenterlines.py`.

Fixes following error when using `vmtkDistanceToCenterlines`:
```python
  File ".../vmtk/master/3j5eq5bfjj7uldmd/lib/python3.8/site-packages/vmtk/vmtkdistancetocenterlines.py", line 108, in Execute
    if (self.UseRadiusThreshold):
AttributeError: 'vmtkDistanceToCenterlines' object has no attribute 'UseRadiusThreshold'
```


lassoan (approved):
Good catch, thanks for the fix!

It would be nice if you could also add a short description of the two parameters in the `self.SetInputMembers...` call.

PR #454, opened 2023-04-19
Update vmtkmeshwriter.py

ramen-newdals:
Changed WriteDolfinMeshFile when writing a compressed mesh gzip in 1.12 does not automatically convert strings to bytes. changing the open mode to wt fixes this issue preventing "Expected a sequence of byes not a string" errors.


lassoan (approved):
Thank you, the change looks good to me.

PR #465, opened 2024-05-09
vmtk WASM Compilation

rodrigobasilio2022:
This PR aims to make all necessary changes to add wasm compilation capabilities to vmtk, so it can be used in web browsers

### DIFF
--- a/vtkVmtk/ComputationalGeometry/vtkvmtkPolyDataCenterlines.cxx
+++ b/vtkVmtk/ComputationalGeometry/vtkvmtkPolyDataCenterlines.cxx
@@ -300,14 +300,26 @@ int vtkvmtkPolyDataCenterlines::RequestData(
     voronoiDiagramFilter->Delete();
   }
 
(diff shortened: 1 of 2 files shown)


lassoan:
Thank you for sharing this. What VTK and ITK version did you use for the build? Probably most changes are for supporting more recent ITK/VTK versions and we may need to add version checks to allow building with both older and newer ITK/VTK versions (or officially drop support for older versions).

rodrigobasilio2022:
Yes the most recent version of both ITK and VTK. I created an Docker file to compile code with both ITK and VTK and I have some instructions in how to compile this with the docker. I can put that in somewhere in the repo. Where do you think is the most suitable place, based on the standards of this repo? There is another problem, the vtkArrayCalculator appears to not be working properly. Need some help in configuration of the build part.

lassoan:
If you still have the instructions then you could add them into a new `Build instructions` section and a `WASMP compilation` subsection in the https://github.com/vmtk/vmtk/edit/master/README.md file. Thank you!

Issue #486, opened 2026-07-15
BUG: SteepestDescentLineTracer corrupts centerline points (unsafe GetPoint aliasing)


DoraemonJack:
`vtkvmtkSteepestDescentLineTracer` (and the same pattern in `vtkvmtkSteepestDescentShooter`) interpolates edge endpoints with:

```cpp
currentPoint[0] = input->GetPoint(currentEdge[0])[0] * (1.0 - currentS)
                + input->GetPoint(currentEdge[1])[0] * currentS;
// same for y, z
```

On VTK 9, when point coordinates are stored as `float`, `double* GetPoint(vtkIdType)` returns a pointer to a **shared temporary buffer**. Calling `GetPoint` twice in the same expression can overwrite the first endpoint before it is read, so the lerp collapses to the second endpoint:

`B*(1-s) + B*s = B`

This produces jagged centerlines (large turning angles) even though the discrete edge sequence / `EdgePCoord` are correct.

VTK documents this limitation and recommends `GetPoint(id, double x[3])` instead.

## Proposed fix

Copy endpoints into locals before interpolating (already used elsewhere in `vtkvmtkNonManifoldSteepestDescent`):

```cpp
double edgePoint0[3], edgePoint1[3];
input->GetPoint(currentEdge[0], edgePoint0);
input->GetPoint(currentEdge[1], edgePoint1);
currentPoint[0] = edgePoint0[0] * (1.0 - currentS) + edgePoint1[0] * currentS;
currentPoint[1] = edgePoint0[1] * (1.0 - currentS) + edgePoint1[1] * currentS;
currentPoint[2] = edgePoint0[2] * (1.0 - currentS) + edgePoint1[2] * currentS;
```

Same change in:

- `vtkVmtk/ComputationalGeometry/vtkvmtkSteepestDescentLineTracer.cxx`
- `vtkVmtk/ComputationalGeometry/vtkvmtkSteepestDescentShooter.cxx`

## Environment

- VMTK: current master
- VTK: 9.1 (also relevant for newer VTK 9.x)
- Reproduced on Windows / MSVC

lassoan:
The original code should work well in most cases, but I agree that there is a risk that the compiler happens to choose an evaluation order that leads to incorrect results. 

Did you test that on your computer the original code produced different result than with the proposed fix?

[...]

PR #487, opened 2026-07-15
fix: avoid unsafe GetPoint aliasing in steepest-descent interpolation

DoraemonJack:
## Summary
Fixes unsafe use of `double* GetPoint(vtkIdType)` twice in one interpolation expression in:
- `vtkvmtkSteepestDescentLineTracer`
- `vtkvmtkSteepestDescentShooter`

Under VTK 9 with float point storage, the returned pointer refers to a shared temporary buffer, so the lerp can collapse to the second endpoint.

## Test plan
- [x] Same mesh + same seeds on VTK 9.1 / MSVC
- [x] Before: jagged centerline (mean turn ~21°, max ~178°)
- [x] After: matches expected smooth path (mean turn ~5°, max ~40°)

Related issue: #486


lassoan (approved):
Thank you, it looks good.

PR #493, opened 2026-07-21
Move static metadata to pyproject.toml (PEP 621)

mwtoews:
This PR moves static metadata from setup.py to pyproject.toml as specified by [PEP 621](https://peps.python.org/pep-0621/) and supported by [setuptools >= 61](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). The aim is to get close to a 1:1 conversion, with a minor change to the licence metadata described next.

The minimum version of setuptools is set to >=77 to support [PEP 639](https://peps.python.org/pep-0639/) to improve licence metadata. This involves specifying a valid SPDX identifier (an expression can also be used, if preferred), and removing the deprecated licence classifier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,11 +6,12 @@ __pycache__
 .cache
 .DS_Store
 .vscode
+*.lock
 # Build directories at the repository root (anchored so that files elsewhere
 # whose names start with "build", e.g. .github/workflows/build-*.yml, are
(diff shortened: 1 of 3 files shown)


lassoan (approved):
Thank you, it looks good to me. I'll merge if CI is all green.

PR #495, opened 2026-07-21
Fix typos

mwtoews:
Fix typos identified and corrected with [codespell](https://github.com/codespell-project/codespell). A configuration is also added, in case the tool needs to be re-run.

Please review for false positives.


lassoan (approved):
I've reviewed all the changes and they look good. Thank you!

PR #504, opened 2026-09-01
ENH: Allow explicit setting cell entity id of each surface cap, across a flow extension

lassoan:
vtkvmtkCapPolyData numbers each cap from the position of the boundary it closes in the list the boundary extractor filter returns. That order is a result of the input's cell and point numbering, so it depends on how the surface happens to be meshed and on which other boundaries exist. However, for some workflows it may be necessary to assign a predetermined cell ID to a cap and this assignment should also work robustly when flow extensions are added.

vtkvmtkCapPolyData takes BoundaryCellEntityIds, indexed consistently with BoundaryIds like everything else in these filters: entry i is the id to write into CellEntityIdsArrayName for the cap of boundary i, and a negative or missing entry keeps the id that boundary's position would give it. A chosen id is used as it stands, with CellEntityIdOffset not added to it, while the offset still lands on the cells copied from the input and on any boundary left to its positional id; the documentation of all three spells that out, since it means an output can carry ids from both rules at once.

Growing a flow extension replaces a boundary with a new one at the tip of the extension, several radii down the branch and made of new points, and the extraction order of the output is not the input's - a boundary left unextended is met before the extensions appended after it. So vtkvmtkPolyDataFlowExtensionsFilter now reports, in GetOutputBoundaryIds(), where each of its input boundaries ended up in its own output. That is the filter's own knowledge: nothing else can relate the two, and a caller that assumed the order was preserved would silently mislabel a face.

Both are exposed to the pypes scripts, so the pair can be used from a pipeline: vmtkflowextensions reports the mapping in outputboundaryids, and vmtksurfacecapper takes the ids in cellentityids for its centerpoint method.

Tests cover the mapping accounting for every boundary and pointing at the boundary that replaced each one, the case where it is not the identity (extending one boundary of two), the capper honouring the ids, falling back where it is given none, and leaving a chosen id free of the offset, and the two together: an id chosen for a vessel end still lands on that end after its boundary has been replaced.

### DIFF
--- a/tests/test_vmtkScripts/test_vmtkflowextensions.py
+++ b/tests/test_vmtkScripts/test_vmtkflowextensions.py
@@ -86,7 +86,7 @@ def boundary_profiles(surface):
     return profiles
 
 
-def extend(surface, preserveshape, centerlines=None, **kwargs):
+def run_extensions(surface, preserveshape, centerlines=None, **kwargs):
     extensions = flowextensions.vmtkFlowExtensions()
     extensions.Surface = surface
     extensions.PreserveCrossSectionShape = preserveshape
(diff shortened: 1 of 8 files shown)


lassoan:
Closing this pull request - will replace it with a solution that makes flow extensions and capping even more robust and simpler.

PR #514, opened 2026-09-17
Add centerline branch splitting for reduced-order models, ported from SimVascular

lassoan:
SimVascular generates reduced-order (0D/1D) blood flow models from centerlines split into branches and bifurcations. It computes those with its own fork of two VMTK classes, `vtkvmtkPolyDataBranchSections` and `vtkvmtkPolyDataCenterlineSections`. This PR brings that processing into VMTK, fixed and cleaned up, so that SimVascular can use the VMTK package from PyPI instead of shipping a patched copy (SimVascular/SimVascular#1193).

The method is part of the framework described in M. R. Pfaller et al., "Automated generation of 0D and 1D reduced-order models of patient-specific blood flow", *Int. J. Numer. Methods Biomed. Eng.* 38(10), e3639, 2022, https://doi.org/10.1002/cnm.3639.

### Commits

**1. Add section helpers from SimVascular to `vtkvmtkPolyDataBranchSections`**

| Method | What it does |
|---|---|
| `ExtractCylinderSection(..., idsArrayName)` | New optional argument. When set, that point data array of the surface (such as branch labels) is copied to the section points from the nearer end point of each cut edge, so the section only has labels that exist on the surface. The input can be any surface, not only a single branch. |
| `CountBranchSectionCenterlines` | Counts the centerlines that cross the section plane inside the section. |
| `CountBranchSectionIds` | Counts the distinct ids on a section. |

The helpers are ported onto the current section extraction rather than copied from the fork, which has these problems:

- **Section extraction:** it rebuilt the section with `vtkStripper` and computed the area with `vtkDelaunay2D`, which left `closed` unset and gave wrong areas for sections not roughly in the xy plane.
- **Labels:** it interpolated point data along the cut edges, which turns labels into values that exist nowhere on the surface (between branches 0 and 3, sections got 1s and 2s), so sections within a branch could be counted as bifurcations.
- **Centerline count:** its projection divided by zero when the plane normal was along z and flattened the section into a line at 45° to z. Here the count uses an orthonormal frame of the section plane and `vtkPolygon::PointInPolygon`.

`ExtractCylinderSection` also now returns an empty section (and `closed` false) when the plane misses the surface, even if the section object is reused, and no longer leaks its plane and cleaner.

**2. Speed up centerline sections and report progress (`vtkvmtkPolyDataCenterlineSections`)**

- **Cutting only crossed cells:** each section is cut from only the surface cells its plane crosses, found from the signed distances of the points, instead of all cells. Cells the plane does not cross contribute nothing to a cut, so sections are unchanged, including contours far from the centerline point.
- **Crossings without `vtkCutter`:** `CountBranchSectionCenterlines` computes where centerline segments cross the plane directly, the same way as `vtkCutter` with `vtkLine::Contour`, instead of running a cutter for every section (whose multithreaded path cost about ten times the serial cut for these small inputs). On 9000 planes, with float and double points, the number of crossings equals the number of points `vtkCutter` outputs in every case.
- **Progress and abort:** progress is reported about every 1% of centerline points, and abort requests are honored (`CheckAbort` on VTK 9.4+, `AbortExecute` otherwise).
- **Subclassing:** the section arrays are created in `ComputeCenterlineSections`, so a subclass with its own `RequestData` only needs to call it.

**3. Add `vtkvmtkPolyDataCenterlineBranchSplitting`, ported from SimVascular**

A subclass of `vtkvmtkPolyDataCenterlineSections`. From a closed surface and centerlines computed from one inlet to each outlet, it:
1. merges the centerlines into a tree, smooths it and refines it next to the caps;
2. cuts a section at every point and classifies it as branch or bifurcation, from `CountBranchSectionCenterlines` and `CountBranchSectionIds`;
3. cleans up the bifurcation regions, numbers branches and bifurcations (`BranchId`, `BifurcationId`) with the distance along each (`Path`);
4. labels the surface with branch and bifurcation ids.

The outputs are the arrays SimVascular's `sv_rom_simulation` and `sv_rom_extract_results` read. Array names are configurable, with SimVascular's names as defaults. Each step reports progress, and abort is checked between steps.

Changes from SimVascular's implementation:

- **Error reporting:** errors go through `vtkErrorMacro` and fail the update; progress printing and debug code are removed.
- **Surface normals:** normals the input surface already has are ignored. SimVascular crashed when a surface had normals under a name other than `Normals`, and `vtkPolyDataNormals` keeps existing normals even when they point into the vessel.
- **VTK 9.4+:** region ids from `vtkConnectivityFilter` are converted to explicit arrays. They are implicit arrays since VTK 9.4, which `vtkAppendFilter` silently drops, so `BranchId`/`BifurcationId` were lost and the final surface labeling failed.
- **Robustness:** cap refinement and point reordering use explicit point id maps instead of index arithmetic that assumed each outlet directly follows its neighbor; point adjacency is computed from the cells instead of `vtkPolyData` links (which form a reference cycle).
- **Leaks:** a `vtkIntArray` leaked on every update; temporary arrays are no longer left on the outputs.

The class documentation credits the SimVascular implementation (Martin R. Pfaller, with changes by Luca Pegolotti and David W. Parker), and the source files include SimVascular's BSD license notice.

### Validation

- **Against SimVascular's code:** SimVascular's two classes were built unchanged (except the `RegionId` conversion needed on VTK 9.5) against the same VTK and run on the same inputs: VMTK's aorta test data, two [Vascular Model Repository](https://www.vascularmodel.com) models (`0159_H_AO_H`, `0187_H_ABAO_AAA`, centerlines computed as SimVascular does) and a synthetic vessel tree. Centerline points, `BranchId`, `BifurcationId`, `Path`, `CenterlineId`, section normals and surface labels are identical, except around 2 sections of `0159_H_AO_H` and 2 of the synthetic tree, where SimVascular counted interpolated labels that exist nowhere on the surface. Section areas, sizes and shapes differ as fixed in commit 1.
- **Against SimVascular's published results:** for `0159_H_AO_H`, the `centerlines.vtp` shipped with the model has the same 5 branches and 1 bifurcation, with branch lengths within 0.11 and mean areas within 1%.
- **Speed:** commit 2 leaves all outputs of branch splitting identical and makes it 3.6–4.2× faster (for example 21 s → 5 s on `0159_H_AO_H`).

### Tests

- `test_vmtkbranchsections.py`: the section helpers on a synthetic tube tilted at several angles, including labels copied without interpolation.
- `test_vmtkcenterlinebranchsplitting.py` (new): the aorta test data and `0159_H_AO_H`, each split once, checking branch and bifurcation structure, lengths and consistency, surface normals and custom array names, progress and abort, and invalid centerlines.
- The `0159_H_AO_H` model was added to vmtk-test-data (vmtk/vmtk-test-data@8025596, with the repository's license terms), and the submodule is updated.

Built against Slicer 5.13 / VTK 9.5 on Windows. `test_vmtkbranchsections.py`, `test_vmtkbifurcationsections.py`, `test_vmtkcenterlinesections.py` and `test_vmtkcenterlinebranchsplitting.py` pass (219 tests in about 10 s, of which branch splitting takes about 8 s); commit 2 alone passes the first three.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lassoan:
The PR is in draft until it is clarified if SimVascular's vtkvmtkPolyDataCenterlineSections changes are needed as well.

